PR #23, opened 2021-09-14
Bunch of has_* tests

smn:
Needed to support rc3 of the floip spec https://floip.gitbook.io/flow-specification/v/drafts%2F1.0.0-rc3/expressions#functions-for-use-within-tests


yaseenisolated (approved):
looks good to me :D

smn:
Realised I still need to fix the handling of the custom `:test_result` return type

smn:
I'm deciding to stay away with the `.match` stuff for now as I think that'll require a relook at the parser + AST setup that I'm unable to take on right now. For now each of these `test` functions returns a boolean and nothing else, which I expect is enough for them to fulfil their test functions in rc3.

PR #25, opened 2021-09-22
Return error tuple

smn:
We're being inconsistent with returning tuples vs throwing errors.
This addresses some of that.

### DIFF
--- a/lib/expression.ex
+++ b/lib/expression.ex
@@ -76,6 +76,9 @@ defmodule Expression do
     with {:ok, ast} <- parse_expression(expression),
          result <- Eval.evaluate!([substitution: ast], context, mod) do
       result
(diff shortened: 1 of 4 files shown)


lib/expression.ex
```diff
@@ -70,6 +70,15 @@ defmodule Expression do
     end
   end
 
+  def evaluate_block!(expression, context \\ %{}, mod \\ Expression.Callbacks)
+
+  def evaluate_block!(expression, context, mod) do
+    with {:ok, ast} <- parse_expression(expression),
```
yaseenisolated, line 76: optional nit: in this case i don't think the with is necessary.

yaseenisolated (approved):

PR #52, opened 2022-05-19
Fix boolean response, not a tuple

smn:
The various boolean has_* callbacks return a tuple, it should just be a boolean return value.


smn:

PR #53, opened 2022-05-25
2nd iteration on parser & eval

smn:
Addresses things we weren't able to do before:

1. nested functions
2. attributes on functions
3. handle email addresses
4. do slicing on lists!


smn:
@santiagocardo apologies, this is a big PR because writing these parsers is simply verbose and I've added a lot of tests.

.github/workflows/elixir.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        elixir: [1.10.4, 1.11.2, 1.12]
+        elixir: [1.12]
```
smn: no need to support the older versions I think, also I've started using the is_struct guard which was introduced in 1.12

lib/expression/callbacks.ex
```diff
@@ -1169,16 +1169,31 @@ defmodule Expression.Callbacks do
...
+      |> Enum.map(fn {_haystack_word, index} -> index end)
 
-    # future match result Enum.join(results, " ")
-    Enum.any?(results)
+    matched_haystack_words = Enum.map(matched_indices, &Enum.at(haystack_words, &1))
+
+    %{
+      "__value__" => Enum.any?(matched_haystack_words),
+      "match" => Enum.join(matched_haystack_words, " ")
+    }
```
smn, line 1195: there's a few other of these `has_*` functions that allow for `.match` attributes -- I'll update those in a separate PR. Used this one as an example / trial.

lib/expression/context.ex
```diff
@@ -46,28 +46,28 @@ defmodule Expression.Context do
...
-  defp evaluate(binary) when is_binary(binary) do
-    case Expression.parse_literal(binary) do
-      {:literal, literal} -> literal
-      {:error, _reason} -> binary
+  defp evaluate!(binary) when is_binary(binary) do
+    case Expression.Parser.literal(binary) do
+      {:ok, [{:literal, literal}], "", _, _, _} -> literal
+      {:error, _reason, _, _, _, _} -> binary
     end
   end
```
santiagocardo:
```elixir
  defp evaluate!(binary) when is_binary(binary) do
    case Expression.Parser.literal(binary) do
      {:ok, [{:literal, literal}], "", _, _, _} -> literal
      {:error, _reason, _, _, _, _} -> binary
    end
  end

  defp evaluate!(value), do: value
```

.github/workflows/elixir.yml
santiagocardo: why not start using 1.13?

lib/expression/callbacks.ex
```diff
@@ -1169,16 +1169,31 @@ defmodule Expression.Callbacks do
...
-    false
+    %{"__value__" => false, "match" => ""}
 
   """
   def has_any_word(_ctx, haystack, words) do
-    {_patterns, results} = search_words(haystack, words)
+    haystack_words = String.split(haystack)
+    haystacks_lowercase = Enum.map(haystack_words, &String.downcase/1)
+    words = String.split(words)
+    words_lowercase = Enum.map(words, &String.downcase/1)
```
santiagocardo:
```suggestion
    words_lowercase = String.split(words) |> Enum.map(&String.downcase/1)
```

santiagocardo (approved):
This is cool! Hope to work on this at some point to understand the code better!

PR #54, opened 2022-05-26
Experiment with `map()` and lambdas

smn:
```
@(map(foo, &([&1, 'Button'])))
```

when context is `%{"foo" => [1, 2, 3]}` results in `[[1, "Button"], [2, "Button"], [3, "Button"]]` as the output.

within the context of the lambda, the `&1` captures are normal variables and so one can use them as arguments for other functions or operations like arithmetic:

```
@(map(foo, &([&1 * 2, 'Button'])))
```

Oh this also introduces lists as a type, bit of a side effect:

```elixir
iex(1)> Expression.evaluate("@([1,2,3])")
{:ok, [[1, 2, 3]]}
```

example use:

```
iex(7)> Expression.evaluate("@(map([1,2,3], &(&1 * 2)))")
{:ok, [[2, 4, 6]]}
```

### DIFF
--- a/lib/expression/callbacks.ex
+++ b/lib/expression/callbacks.ex
@@ -1949,4 +1949,12 @@ defmodule Expression.Callbacks do
       _ -> false
     end
   end
+
+  def map(_ctx, enumerable, mapper) do
+    enumerable
+    # wrap in a list to be passed as a list of arguments
+    |> Enum.map(&[&1])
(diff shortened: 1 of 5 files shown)


smn:
This PR builds on #53

smn:
I’ll address the test failures. Due to renaming list to :access and introducing proper list support.

santiagocardo (approved):
Brilliant! ✨

PR #55, opened 2022-05-28
Add as boolean!

smn:
For some use cases, things just need to return a boolean or crash.


santiagocardo (approved):
Thanks @smn, This is looking good to me!

Just a final suggestion: What do you think of renaming these functions to `Expression.evaluate_as_string!` and `Expression.evaluate_as_boolean!`? that way we are more explicit by saying this is going to evaluate and then return a string/boolean

PR #57, opened 2022-05-30
Allow trailing full stops

smn:
Previously `Hello @contact.name.` would break because of the trailing full stop. 
This PR fixes that.


smn:
@neelkes FYI

smn:
@santiagocardo going ahead and merging this to unblock work for a client

PR #58, opened 2022-06-01
Fix key access

smn:
`@IF(record[foo], record[foo], "")` was failing because the `[` and `]` access keys weren't accepted in function arguments.

### DIFF
--- a/lib/expression.ex
+++ b/lib/expression.ex
@@ -75,12 +75,15 @@ defmodule Expression do
     expression
     |> parse!
     |> Eval.eval!(Context.new(context), mod)
+    |> Enum.map(&default_value/1)
   end
 
   def evaluate_as_string!(expression, context \\ %{}, mod \\ Expression.Callbacks) do
     expression
     |> parse!
-    |> Eval.as_string!(Context.new(context), mod)
+    |> Eval.eval!(Context.new(context), mod)
+    |> Enum.map(&default_value/1)
+    |> Enum.map_join(&Kernel.to_string/1)
   end
 
   def evaluate_as_boolean!(expression, context \\ %{}, mod \\ Expression.Callbacks) do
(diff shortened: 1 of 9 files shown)


smn:
😅

lib/expression.ex
```diff
@@ -93,6 +96,10 @@ defmodule Expression do
     end
   end
 
+  defp default_value(%{"__value__" => default_value}), do: default_value
+  defp default_value({:not_found, attributes}), do: "@#{Enum.join(attributes, ".")}"
+  defp default_value(value), do: value
```
smn, line 101: moved this to the top-level because inside `Expression.Eval.eval!` we need to work with the raw values

lib/expression/eval.ex
```diff
@@ -84,9 +85,10 @@ defmodule Expression.Eval do
     ast
     |> Enum.reduce([], &[eval!(&1, context, mod) | &2])
     |> Enum.reverse()
+    |> Enum.map(&not_founds_as_nil/1)
```
smn, line 86: otherwise the `[not_found: ["foo", "bar"]]` is supplied as a truthy argument to functions like `@IF()` and those would then fail

lib/expression/eval.ex
```diff
@@ -33,17 +33,17 @@ defmodule Expression.Eval do
...
-  def eval!({:attribute, [{:atom, subject}, {:atom, key}]}, context, _mod) do
-    get_in(context, [subject, key]) || "@#{subject}.#{key}"
-  end
-
-  def eval!({:attribute, [subject_ast, {:atom, key}]}, context, mod) do
-    subject = eval!(subject_ast, context, mod)
-    get_in(subject, [key])
+  def eval!({:attribute, ast}, context, mod) do
+    Enum.reduce(ast, context, &eval!(&1, &2, mod))
+    # eval!({:attribute, ast}, context, mod)
```
santiagocardo: remove comment

lib/expression/eval.ex
```diff
@@ -33,17 +33,17 @@ defmodule Expression.Eval do
     eval!(ast, context, mod)
   end
 
+  def eval!({:atom, atom}, {:not_found, history}, _mod),
+    do: {:not_found, history ++ [atom]}
+
   def eval!({:atom, atom}, context, _mod) do
-    get_in(context, [atom])
+    # get_in(context, [atom])
```
santiagocardo: remove comment

santiagocardo (approved):
there are some things I'm not following yet but it looks good to go based on what I see from the tests 👍🏼

PR #59, opened 2022-06-02
Fix attribute literal keys

smn:
When we receive a key for an attribute, at times this could be a literal.
The assumption is that all attributes are going to be string based so if we receive
`@foo.123.bar`, `123` will be parsed as a literal but the assumption is that the
context will look like:

```elixir
%{"foo" => %{
  "123" => %{   <--- notice the string key here
    "bar" => "the value"
  }
}}
```


smn:
github having issues, merging despite 1 test check failure.

PR #63, opened 2022-06-23
Fix eager loading

smn:


test/expression/eval_test.exs
```diff
@@ -27,41 +27,49 @@ defmodule Expression.EvalTest do
...
+  end
+
+  test "if" do
+    {:ok, ast, "", _, _, _} =
+      Parser.parse("@if(image_response.status == 200,\nimage_response.body.id,\nfalse)")
+
+    assert false ==
+             Eval.eval!(ast, %{
+               "image_response" => %{"status" => 500, "body" => "Internal Server Error"}
+             })
```
smn, line 40: @santiagocardo all this work just to make sure this doesn't crash ...

smn:
https://github.com/turnhub/flow_runner/pull/47 and https://github.com/turnhub/stacks_dsl/pull/23 use these changes, once released & publish I'll update the mix.exs in those PRs.

lib/expression/callbacks.ex
```diff
@@ -991,134 +928,147 @@ defmodule Expression.Callbacks do
...
-  def word_slice(_ctx, binary, start, stop, by_spaces) when stop < 0 do
+  def word_slice(ctx, binary, start, stop, by_spaces) do
+    [binary, start, stop, by_spaces] = eval_args!([binary, start, stop, by_spaces], ctx)
     splitter = if(by_spaces, do: " ", else: @punctuation_pattern)
 
-    binary
-    |> String.split(splitter)
-    |> Enum.slice((start - 1)..(stop - 1))
-    |> Enum.join(" ")
+    cond do
```
santiagocardo: not that is expensive in this situation, but we can use `case` instead so we keep using conventions in favor of more performant code.

Here and the other places where this is being used

lib/expression/callbacks.ex
```diff
@@ -991,134 +928,147 @@ defmodule Expression.Callbacks do
...
-      iex> Expression.Callbacks.isnumber(%{}, "a")
+      iex> Expression.evaluate!("@isnumber(\\"a\\")")
       false
 
   """
-  def isnumber(_ctx, var) when is_float(var) or is_integer(var), do: true
+  def isnumber(ctx, var) do
+    var = eval!(var, ctx)
+
+    cond do
```
santiagocardo: same here

lib/expression/callbacks.ex
```diff
@@ -991,134 +928,147 @@ defmodule Expression.Callbacks do
...
       false
-      iex> Expression.Callbacks.isstring(%{}, 1)
+      iex> Expression.evaluate!("@isstring(1)")
       false
-      iex> Expression.Callbacks.isstring(%{}, Decimal.new("1.0"))
+      iex> Expression.evaluate!("@isstring(d)", %{"d" => Decimal.new("1.0")})
       false
   """
-  def isstring(_ctx, binary), do: is_binary(binary)
+  def isstring(ctx, binary), do: is_binary(eval!(binary, ctx))
```
santiagocardo, line 1081: can we rename this to `string?` or `is_string`

lib/expression/callbacks.ex
```diff
@@ -21,6 +21,14 @@ defmodule Expression.Callbacks do
   underscore.
   """
 
+  def eval!(ast, ctx) do
+    ast
+    |> Expression.Eval.eval!(ctx, __MODULE__)
+    |> Expression.Eval.not_founds_as_nil()
+  end
+
+  def eval_args!(args, ctx), do: Enum.map(args, &eval!(&1, ctx))
```
santiagocardo: could you please add some docs for these functions?

santiagocardo (approved):
LGTM. huge job, @smn!

lib/expression/callbacks.ex
santiagocardo: omit it if this is just following the `Expressions` specs

smn: the FLOIP spec standard dictates these function names ... 🤷

lib/expression/callbacks.ex
smn: done!

lib/expression/callbacks.ex
smn: done!

lib/expression/callbacks.ex
smn: done!

PR #64, opened 2022-06-28
Stringify primitives properly

smn:
Expressions know of a few primitives but they'd break when we tried to display them as strings.


santiagocardo (approved):
LGTM

PR #72, opened 2022-09-20
fix decimal arithmetic

smn:
Turns out we suck at arithmetic with decimal & floating points

### DIFF
--- a/test/expression/eval_test.exs
+++ b/test/expression/eval_test.exs
@@ -58,7 +58,7 @@ defmodule Expression.EvalTest do
              ] == Eval.eval!(ast, %{})
     end
 
-    test "with arithmatic" do
+    test "with arithmetic" do
       {:ok, ast, "", _, _, _} = Parser.parse("@(map(foo, &([&1, 'Button'])))")
 
       assert [[1, "Button"], [2, "Button"], [3, "Button"]] ==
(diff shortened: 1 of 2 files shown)


test/expression/eval_test.exs
```diff
@@ -139,6 +139,19 @@ defmodule Expression.EvalTest do
...
 
+  test "arithmetic with decimals" do
+    {:ok, ast, "", _, _, _} = Parser.parse("@(1.5 + 1.5)")
+    assert Decimal.new("3.0") == Eval.eval!(ast, %{})
+    {:ok, ast, "", _, _, _} = Parser.parse("@(1.5 + 2.5 * 3.5)")
+    assert Decimal.new("10.25") == Eval.eval!(ast, %{})
+    {:ok, ast, "", _, _, _} = Parser.parse("@((1.5 + 2.5) * 3.5)")
+    assert Decimal.new("14.00") == Eval.eval!(ast, %{})
+    {:ok, ast, "", _, _, _} = Parser.parse("@((1.5 + 2.5) / 3.5)")
+    assert Decimal.new("1.14286") == Decimal.round(Eval.eval!(ast, %{}), 5)
```
smn, line 150: the rounding here is because depending the mood of the cpu it'll print a decimal with different precisions

santiagocardo (approved):
Really nice! Good guarding clauses!

PR #79, opened 2022-10-20
[sc-14778] calling functions across modules

smn:
Calling functions across modules fails in certain scenarios:

If `echo/1` is defined in a custom callback function, it works fine in isolation.
If it's called from within a function defined in the standard function callbacks, it'll fail because of how we're doing function lookups. 

This works: 
```elixir
foo = echo("foo")
text("Hello @proper(foo)")
```

This does not:
```elixir
text("Hello @proper(echo(\"foo\"))")
```

It'll fail because `echo/1` isn't defined in the same module as `proper/1` is .


smn:
recheck

PR #80, opened 2022-10-23
Fix custom callbacks module

santiagocardo:


smn:
I'll update the CLA assistant to exclude our email addresses

santiagocardo:
@smn I made it work, but lambdas are not working correctly because they are making the `EvalHelpers` use `handle/4` as if they were the main `Callbacks` module. It's just one test that is failing :/

santiagocardo:
> @smn I made it work, but lambdas are not working correctly because they are making the `EvalHelpers` use `handle/4` as if they were the main `Callbacks` module. It's just one test that is failing :/

No worries @smn I've fixed it!

lib/expression/callbacks.ex
```diff
@@ -21,81 +21,83 @@ defmodule Expression.Callbacks do
...
   end
 
-  def function_name_helpers do
-    quote do
-      @reserved_words ~w[and if or not]
+  def implements(module \\ Standard, function_name, arguments) do
+    exact_function_name = atom_function_name(function_name)
+    vargs_function_name = atom_function_name("#{function_name}_vargs")
+
+    Code.ensure_loaded!(Standard)
```
smn, line 70: do we need to document what's going on here?

smn: Actually I’m not even sure why this is needed yet

lib/expression/eval.ex
```diff
@@ -60,8 +60,17 @@ defmodule Expression.Eval do
...
     name = opts[:name] || raise "Functions need a name"
-    arguments = opts[:args] || []
-    IO.inspect(mod, label: "mod")
+    args = opts[:args] || []
+
+    arguments =
+      Enum.map(args, fn
+        {type, _args} = function when type in [:function, :lambda] ->
+          value = eval!(function, context, mod)
+          [literal: value]
```
smn, line 69: Can you talk me through why this special casing is needed here?

smn (approved):
I’m happy with the result but I’m still puzzling a bit about how it works :) could you add some more detail perhaps with some code comments ?

lib/expression/callbacks.ex
smn: oh wait maybe I understand now, otherwise the function exported function doesn't work?

smn: I think we can move this to the macro on line 96 to prevent it being called on every function call/evaluation?

lib/expression/eval.ex
```diff
@@ -60,8 +60,17 @@ defmodule Expression.Eval do
...
+
+        argument ->
+          argument
```
smn, line 72:
```suggestion
        arguments ->
          arguments
```

lib/expression/eval.ex
smn: My guess is that we're doing this to force the evaluation of all function arguments scoped to the local module first before attempting to delegate to any other module?

smn: what will this do with something like the `if()` function that does lazy evaluation on arguments?

smn: hm I think our tests already cover for that.

smn:
ok read it again, happy with this but I would suggest we mode the `Code.ensure_loaded/1` call to the macro instead, there's no need to do that on every function call.

smn:
I'll merge this into my PR now

PR #81, opened 2022-10-23
date comparison fixes

smn:
1. We're not including `.match` which is useful to have
2. We're not accepting datestructs to some of these helpers, they're expecting strings.

This PR fixes those.

### DIFF
--- a/lib/expression/callbacks.ex
+++ b/lib/expression/callbacks.ex
@@ -138,6 +138,68 @@ defmodule Expression.Callbacks do
   end
 
   @doc """
+  Calculates a new datetime based on the offset and unit provided.
+
+  The unit can be any of the following values:
+
+  * "Y" for years
+  * "M" for months
+  * "W" for weeks
+  * "D" for days
+  * "h" for hours
+  * "m" for minutes
+  * "s" for seconds
+
+  Specifying a negative offset results in date calculations back in time.
+
+  # Example
+
+      iex> Expression.evaluate!("@datetime_add(date(2022, 11, 1), 1, \\"Y\\")")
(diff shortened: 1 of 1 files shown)


lib/expression/callbacks.ex
```diff
@@ -1214,13 +1282,30 @@ defmodule Expression.Callbacks do
...
     [expression, date_string] = eval_args!([expression, date_string], ctx)
     found_date = extract_dateish(expression)
     test_date = extract_dateish(date_string)
-    # Future match result: found_date
-    found_date == test_date
+
+    result =
+      case found_date do
+        found_date when is_struct(found_date, DateTime) ->
+          DateTime.compare(found_date, test_date) == :eq
```
fedme, line 1299: What if `test_date` is not a `DateTime` struct? Could that happen?

lib/expression/callbacks.ex
```diff
@@ -1214,13 +1282,30 @@ defmodule Expression.Callbacks do
...
+
+        found_date when is_struct(found_date, Date) ->
+          Date.compare(found_date, test_date) == :eq
```
fedme, line 1302: What if `test_date` is not a `Date` struct? Could that happen?

fedme (approved):
LGTM, left a question in a comment

lib/expression/callbacks.ex
smn: Ah no you’re right. Even the date that function returns a date time. I’ll fix.

smn: actually, no, today() returns a date according to the spec.

PR #84, opened 2022-10-29
Auto generate docs samples + doc tests

smn:
It's useful to be able to extract samples from the documentation in a structured format to:

1. Automatically generate tests for them
2. Make them available in a convenient format for use for tooling purposes

Adds the following to any existing doc string already added. Adds to as is if no docstring is present.

Screenshot 2022-10-30 at 11 12 11

Complex types (return values with attributes) are now also automatically documented:

Screenshot 2022-10-31 at 15 47 42

The expressions documented are tested as doctests.

Adds the `expression_docs` function which returns this:

```elixir
iex(2)> Expression.Callbacks.Standard.expression_docs
[
  {"date", ["year", "month", "day"],
   [
     %{
       context: %{"day" => 31, "month" => 1, "year" => 2022},
       doc: "Construct a date from year, month, and day integers",
       expression: "date(year, month, day)",
       result: "2022-01-31T00:00:00Z"
     }
   ]}
]
iex(3)> 
```

The format is:

```elixir
@expression_doc doc: "Construct a date from year, month, and day integers",
                expression: "date(year, month, day)",
                context: %{"year" => 2022, "month" => 1, "day" => 31},
                result: "2022-01-31T00:00:00Z"
```

Where: 
* `doc` is the explanatory text added to the doctest.
* `expression` is the expression we want to test
* `context` is the context the expression is tested against
* `result` is the result we're expecting to get and are asserting against

### DIFF
--- a/lib/expression/autodoc.ex
+++ b/lib/expression/autodoc.ex
@@ -76,7 +76,7 @@ defmodule Expression.Autodoc do
         When used as a Stack expression it returns a #{format_result(expression_doc[:result])}#{format_context(expression_doc[:context])}.
 
         ```
-        #{expression_doc[:expression]}
+        #{expression_doc[:expression]} -> #{inspect(expression_doc[:result])}
         ```
 
         When used as an expression in text, prepend it with an `@`:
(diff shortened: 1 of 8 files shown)


lib/expression/autodoc.ex
```diff
@@ -0,0 +1,204 @@
...
+  """
+  defmacro __using__(_args) do
+    quote do
+      @expression_docs []
+      Module.register_attribute(__MODULE__, :expression_doc, accumulate: true)
+      @on_definition Expression.Autodoc
+      @before_compile Expression.Autodoc
+
+      import Expression.Autodoc
+      require Expression.Autodoc
```
santiagocardo, line 35: do we need both of these?

santiagocardo: do we need both of these?

santiagocardo (approved):
I like this approach! Let's explore building around this!

PR #88, opened 2022-12-07
fix comparisons against default values

smn:
Complex values (values returning a Map with a `__value__` key) are failing when eval'd against simple operators like `==`


smn:
Screenshot 2022-12-08 at 10 02 19

Github is having issues but tests pass

PR #89, opened 2022-12-13
Autodoc should reference callback module

smn:
Otherwise autodoc stuff we write for other callback modules fail because it attempts to test callback expressions against the standard module

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -16,11 +16,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Set up Elixir
-        uses: actions/setup-elixir@v1
+        uses: erlef/setup-beam@v1
         with:
           elixir-version: ${{ matrix.elixir }}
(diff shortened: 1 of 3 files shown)


.github/workflows/elixir.yml
```diff
@@ -10,17 +10,16 @@ jobs:
...
-        elixir: [1.13]
-        otp: [24]
+        elixir: ["1.14.0"]
+        otp: ["25"]
 
     steps:
       - uses: actions/checkout@v2
       - name: Set up Elixir
-        uses: actions/setup-elixir@v1
+        uses: erlef/setup-beam@v1
```
santiagocardo, line 19: is this like a stable version for CI?

santiagocardo (approved):
Nice! LGTM

.github/workflows/elixir.yml
smn: yes! the old one is deprecated

PR #93, opened 2023-01-25
Allow underscores in integer literals

smn:
Aligning with Elixir conventions of allowing `123_456` as a more readable means of displaying an Integer

### DIFF
--- a/lib/literal_helpers.ex
+++ b/lib/literal_helpers.ex
@@ -4,7 +4,13 @@ defmodule Expression.LiteralHelpers do
 
   def int do
     optional(string("-"))
-    |> concat(integer(min: 1))
+    |> times(
+      choice([
+        integer(min: 1),
+        ignore(utf8_char([?_]))
+      ]),
+      min: 1
+    )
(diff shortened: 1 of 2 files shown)


lib/literal_helpers.ex
```diff
@@ -4,7 +4,13 @@ defmodule Expression.LiteralHelpers do
 
   def int do
     optional(string("-"))
-    |> concat(integer(min: 1))
+    |> times(
+      choice([
```
fedme, line 8: what's `choice`?

fedme (approved):

lib/literal_helpers.ex
smn: it means the parser *must* choose one of the two options, either we hit a integer, or we hit an underscore (which we then ignore).

The choice function is documented here https://hexdocs.pm/nimble_parsec/NimbleParsec.html#choice/2
The times function is documented here https://hexdocs.pm/nimble_parsec/NimbleParsec.html#times/3

PR #95, opened 2023-02-07
do not strip leading zeros in floats & decimals

smn:
We parse a decimal by parsing two ints separated by a full stop.
However reading `0000001` as an int results in `1`, should read it as a string first and then convert it, otherwise '0.0001` is read as `0.1`


smn:
This still needs a test. Will add.

PR #96, opened 2023-02-07
create datetime helpers to aid with stack scheduling

JuliaMathias:
For our scheduled stacks functionality we will also create a datetime helper to make scheduling more pleasant. Something like: `next("monday", "08:00")`, or even `next("monday", "08:00", previous_appointment_in_datetime)` that would return the correct date for the next schedule in datetime.

It would then be used like this: `schedule_stack("uuid", at: next("monday", "08:00", previous_appointment)`


JuliaMathias:

lib/timezone_helpers.ex
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  defmacro __using__(_opts) do
+    quote do
+      @before_compile Expression.TimezoneHelpers
+    end
+  end
+
+  def all(), do: @map_entries
+end
```
JuliaMathias: I pretty much just copied this from engage. I'm not sure if there was a better way to get the timezone of the user, but this was the solution I found.

lib/expression/callbacks/standard.ex
```diff
@@ -100,6 +114,77 @@ defmodule Expression.Callbacks.Standard do
...
+  If no base date is supplied, it will fallback to current time.
+  """
+  @expression_doc doc:
+                    "Shifts a datetime to the next occurrence of a given day of the week and a set time.",
+                  expression: "next(\"monday\", \"10:00\", datetime)",
+                  context: %{
+                    "datetime" => ~U[2023-02-03 20:18:03Z],
+                    "number" => %{"from_addr" => "+5521987724322"}
+                  },
+                  fake_result: Map.merge(@datetime_struct, %{day: 6, hour: 10, minute: 0})
```
JuliaMathias, line 129: I did it this way because `DateTime.new()` doesn't work here, I get a compilation error. I thought that having a fake result would still be pretty useful to get the automatically generated docs.

config/config.exs
```diff
@@ -0,0 +1,3 @@
+import Config
+
+config :elixir, :time_zone_database, Tzdata.TimeZoneDatabase
```
JuliaMathias, line 3: this was necessary to use Datetimes with timezones.

lib/expression/callbacks/standard.ex
```diff
@@ -26,6 +26,20 @@ defmodule Expression.Callbacks.Standard do
...
+  @datetime_struct %DateTime{
+    calendar: Calendar.ISO,
+    day: nil,
+    hour: nil,
+    microsecond: {0, 0},
+    minute: nil,
+    month: 2,
+    second: 0,
+    std_offset: 0,
+    time_zone: "America/Sao_Paulo",
```
smn, line 38: ❤️

lib/expression/callbacks/standard.ex
```diff
@@ -100,6 +114,74 @@ defmodule Expression.Callbacks.Standard do
...
+                    "If the day of the week is the same as the base date, the next occurrence is 7 days later.",
+                  expression: "next(\"thursday\", \"10:30\", datetime)",
+                  context: %{
+                    "datetime" => ~U[2023-02-02 20:18:03Z],
+                    "number" => %{"from_addr" => "+5521987724322"}
+                  },
+                  fake_result: Map.merge(@datetime_struct, %{day: 9, hour: 10, minute: 30})
+  @spec next(map, {:literal, String.t()}, {:literal, String.t()}, any) :: DateTime.t()
+  def next(
+        %{"number" => %{"from_addr" => from_addr}} = ctx,
```
smn, line 140: This is a bit of philosophical question. When we do `next(monday, "10:30", now())` do we anchor on the timezone of the number of the service or on the timezone of the number of the recipient? My gut feel is that it should probably be the number of the recipient?

lib/expression/callbacks/standard.ex
```diff
@@ -100,6 +114,74 @@ defmodule Expression.Callbacks.Standard do
...
+        desired_day,
+        time,
+        base_date \\ DateTime.utc_now()
+      ) do
+    base_date =
+      if is_struct(base_date) == false,
+        do: extract_datetimeish(eval!(base_date, ctx)),
+        else: base_date
```
smn, line 148: yeah this is a bit of an odd situation and I've not come to a good design decision on how to handle default arguments in callback functions.

Explicit arguments supplied are AST and they need to be evaluated before being useful, however, default arguments are not AST and are ready to be used. That's why you're doing the `eval!()` here. The places I've come across this I've just defined two separate functions, in this case it would be a `next/3` and a `next/4` explicitly. 

Do you have any better suggestions for handling this?

lib/expression/callbacks/standard.ex
```diff
@@ -100,6 +114,74 @@ defmodule Expression.Callbacks.Standard do
...
+
+    [desired_day, time] = eval_args!([desired_day, time], ctx)
+
+    [hour, minute] = time |> String.split(":") |> Enum.map(&String.to_integer/1)
```
smn, line 152: There are some parsers in [Expression.DateHelpers](https://github.com/turnhub/expression/blob/446f3a352fe3d3a4335aea4ef907d22bf4eb5e01/lib/date_helpers.ex#L79) that parse a datetime. Those are using an internal parser called `iso_time()` and `us_time()` -- maybe we can split those into into separate things and parse `10:30` into a %Time{} straight away?

lib/expression/callbacks/standard.ex
smn: Maybe these no local return things that Dialyzer is complaining about relate to this?

lib/expression/callbacks/standard.ex
smn: Happy to treat this as scope creep though! it's up to you. The current solution works fine as well.

lib/expression/callbacks/standard.ex
smn: Also `number.from_addr` is quite Turn specific. In FLOIP land, the contact's details would be under the `%{"contact" => %{...}}`. I'd need to check what the expected key of the address of the contact is.

JuliaMathias: Yeah, I guess it would be better to get the recipients' number. What would be the correct field for that? @smn

lib/expression/callbacks/standard.ex
JuliaMathias: At first I did something like that, but for the Timex.set function I need the hour and minutes to be separated, so putting them into a time struct doesn't work.

smn:
@JuliaMathias I think we can remove the datetime_parser dependency if that helps unblock this.

test/expression_callbacks_test.exs
```diff
@@ -1,5 +1,48 @@
...
   doctest Expression.Callbacks.Standard, import: true
+
+  @context %{
+    "base_date" => ~U[2023-02-03 20:18:03Z],
+    "number" => %{"from_addr" => "+552197295926"}
+  }
+
+  describe "next/4" do
+    test "Shifts a datetime to the next occurrence of a given day of the week and a set time." do
+      assert Expression.evaluate_block!(~s(next\(monday, 10:00, base_date\)), @context) ==
```
smn, line 14: How is this working when `monday` here is not a quoted string? 
My guess is that it should be something like
```elixir
Expression.evaluate_block(~s[next("monday", "10:00", base_date)], ctx) 
```

lib/expression/callbacks/standard.ex
smn: According to the the FLOIP spec it's https://floip.gitbook.io/flow-specification/context#contact `contact.phonenumber` but I don't think we set that yet in Turn.

Our options are either to introduce that field into Turn (it's currently in contact.whatsapp_id in Turn) or moving this helper to the Turn callbacks module (outside of this repo) so that we make it easier to keep track of customisations & standard deviations in a bit more centralised fashion in the Turn codebase.

lib/expression/callbacks/standard.ex
smn: I tackled this here https://github.com/turnhub/expression/pull/97

lib/expression/callbacks/standard.ex
smn: Moving this to the custom callbacks function in Turn also has the benefit of not needing to copy the phone number prefix stuff below.

JuliaMathias: I didn't know that was even an option. But yeah, I think I'm gonna do that @smn

test/expression_callbacks_test.exs
JuliaMathias: it wasn't working. I had changed it but not pushed yet

PR #100, opened 2023-02-18
revisit parser & eval

smn:


lib/expression/v2/eval.ex
```diff
@@ -0,0 +1,126 @@
...
+  available in the callback module specified.
+
+  It does this by emitting valid Elixir AST, mimicking what `quote/2` does.
+
+  The Elixir AST is then supplied to `Code.eval_quoted_with_env/3` with the binding
+  and Elixir evaluates it for us, reducing the need for us to ensure that our 
+  eval is correct and instead relying on us to generate correct AST instead.
+
+  The callback module referenced is inserted as the module for any function
+  that is called. So if we an expression uses a function called `foo(1, 2, 3)`
```
santiagocardo:
```suggestion
  that is called. So if an expression uses a function called `foo(1, 2, 3)`
```

lib/expression/v2/eval.ex
```diff
@@ -0,0 +1,126 @@
...
+  then the callback module **must** provide a `foo/3` function.
+
+  There is some special handling of some functions that either require lazy
```
santiagocardo:
```suggestion
  There is some special handling of some functions that either requires lazy
```

lib/expression/v2/eval.ex
```diff
@@ -0,0 +1,126 @@
...
+    term
+  end
+
+  @doc """
+  Convert the AST returned from `Expression.V2.parse/1` into valid Elixir AST
+  that can be used by `Code.eval_quoted_with_env/3`.
+
+  There is some special handling here:
+
+  1. Lists are recursed into to ensure that all list items are properly quoted.
```
santiagocardo:
```suggestion
  1. Lists are recursed to ensure that all list items are properly quoted.
```

santiagocardo (commented):
IHM, this is looking great! I really like the approach of relying on Elixir to evaluate things for us, and instead, we focus on generating a correct AST.

santiagocardo:
I didn't fully understand the parser as it uses a lot of **NimbleParsec**'s DSL. I think it's a matter of studying the lib, but from what I can see in tests, it's working nicely!

smn:
@santiagocardo yeah the NimbleParsec DSL would require some reading into those' docs. Not sure how better to document that here...

lib/expression/v2/eval.ex
```diff
@@ -23,16 +19,24 @@ defmodule Expression.V2.Eval do
   MyProject.Callbacks.callback(context, "foo", [1, 2, 3])
   ```
 
-  There is some special handling of some functions that either requires lazy
-  loading or have specific Elixir AST syntax requirements.
+  There is some special handling of some functions have specific Elixir AST
```
JuliaMathias:
```suggestion
  There is some special handling of some functions that have specific Elixir AST 
```

benchmark.exs
```diff
@@ -0,0 +1,19 @@
...
+      "Hello @contact.name, you were born in @(YEAR(contact.birthday))",
+      %{"contact" => %{"name" => "mary", "birthday" => ~U[1920-02-02T00:00:00+0000]}}
+    },
+    "map" => {~S|hi @map(0..10, &([&1, concatenate("Button ", &1)])) there|, %{}},
+    "if" =>
+      {"yebo @if(something.false, 1, contact.bar) yes",
+       %{"something" => %{"false" => false}, "contact" => %{}}},
+    "arithmetic" => {"3 * (5 + 2) = @(3 * (5 + 2))", %{}}
+  }
+)
```
JuliaMathias, line 19: I don't think it's a huge problem but did you find out why V2 is slower?

lib/expression/v2/callbacks.ex
```diff
@@ -0,0 +1,52 @@
+defmodule Expression.V2.Callbacks do
+  @moduledoc """
+  Callback functions to be used in Expressions.
+
+  This is the same idea as `Expression.Callbacks.Standard` but 
+  it's in a rough shape, mostly to just prove that this all works.
```
JuliaMathias: This made me think about what would be the plan to "migrate" from v1 to v2. Also would we have to change anything for our custom callbacks?

lib/expression/v2/compile.ex
```diff
@@ -0,0 +1,205 @@
...
+
+  Any function calls are applied to the callback module referenced in the context.
+  So if an expression uses a function called `foo(1, 2, 3)` then the callback's 
+  `callback/3` function will be called as follows:
+
+  ```elixir
+  apply(context.callback_module, :callback, ["foo", [1, 2, 3]])
+  ```
+
+  There is some special handling of some functions have specific Elixir AST
```
JuliaMathias:
```suggestion
  There is some special handling of some functions that have specific Elixir AST 
```

lib/expression/v2/compile.ex
```diff
@@ -0,0 +1,205 @@
...
+       quoted(a),
+       quoted(b)
+     ]}
+  end
+
+  defp quoted({:__attribute__, [a, b]}) do
+    # For any other attributes, we're assuming we just want to read
+    # a property off of a Map so
+    # `foo[bar]` becomes `Map.get(foo, bar)`
+    # `foo["bar"]` becomse `Map.get(foo, "bar")` etc
```
JuliaMathias:
```suggestion
    # `foo["bar"]` becomes `Map.get(foo, "bar")` etc
```

lib/expression/v2/parser.ex
```diff
@@ -0,0 +1,405 @@
...
+    |> replace(false)
+
+  boolean =
+    choice([
+      boolean_true,
+      boolean_false
+    ])
+
+  # These are just regular floats, previous iteration used the
+  # Decimal library but that just made some simple arithmatic
```
JuliaMathias:
```suggestion
  # Decimal library but that just made some simple arithmetic
```

lib/expression/v2/parser.ex
```diff
@@ -0,0 +1,405 @@
...
+      |> ignore(string("]")),
+      min: 1
+    )
+
+  # The following operators determine the order of operations, as per
+  # https://en.wikipedia.org/wiki/Order_of_operations
+
+  # Normally this would also have root but we don't have a shorthand for that
+  # and so rather than a list of options, this is just the exponent operator.
+  # This has higher precendence.
```
JuliaMathias:
```suggestion
  # This has higher precedence.
```

lib/expression/v2/parser.ex
```diff
@@ -0,0 +1,405 @@
...
+  exponentiation_operator = string("^")
+
+  # Multiplation & division is second
```
JuliaMathias:
```suggestion
  # Multiplication & division is second
```

lib/expression/v2/parser.ex
```diff
@@ -0,0 +1,405 @@
...
+        label(atom, "an atom")
+      ]),
+      min: 1
+    )
+
+  # Below are the precendence parsers, each gives the higher precendence
+  # a change to parse its things _before_ it itself attempts to do so.
+  # This is how the precendence is guaranteed.
+
+  # First operator precendence parser
```
JuliaMathias:
```suggestion
  # Below are the precedence parsers, each gives the higher precedence
  # a change to parse its things _before_ it itself attempts to do so.
  # This is how the precedence is guaranteed.

  # First operator precedence parser
```

lib/expression/v2/parser.ex
```diff
@@ -0,0 +1,405 @@
...
+    |> repeat(
+      exponentiation_operator
+      |> label("an operator")
+      |> ignore_surrounding_whitespace.()
+      |> concat(term)
+    )
+    |> reduce(:fold_infixl)
+  )
+
+  # Second operator precendence parser
```
JuliaMathias:
```suggestion
  # Second operator precedence parser
```

lib/expression/v2/parser.ex
```diff
@@ -0,0 +1,405 @@
...
+    parsec(:exponentiation)
+    |> repeat(
+      multiplication_division_operator
+      |> ignore_surrounding_whitespace.()
+      |> concat(parsec(:exponentiation))
+    )
+    |> reduce(:fold_infixl)
+  )
+
+  # Third operator precendence parser
```
JuliaMathias:
```suggestion
  # Third operator precedence parser
```

test/expression/v2/eval_test.exs
```diff
@@ -0,0 +1,108 @@
...
+
+    test "indices on lists" do
+      assert "qux" == eval("foo[2]", %{"foo" => [1, 2, "qux"]})
+    end
+
+    test "function calls" do
+      assert 1 == eval("echo(1)")
+    end
+
+    test "arithmatic" do
```
JuliaMathias:
```suggestion
    test "arithmetic" do
```

test/expression/v2/parser_test.exs
```diff
@@ -0,0 +1,211 @@
...
+      assert {:ok, [[1.123456789]], "", _, _, _} = Parser.parse("@(1.1234567890)")
+    end
+
+    test "atom" do
+      assert {:ok, [["foo"]], "", _, _, _} = Parser.parse("@(foo)")
+    end
+  end
+
+  describe "functions" do
+    test "expresssion/1" do
```
JuliaMathias:
```suggestion
    test "expression/1" do
```

test/expression/v2/parser_test.exs
```diff
@@ -0,0 +1,211 @@
...
+      assert {:ok, ["hi ", [{"now", []}]], "", _, _, _} = Parser.parse("hi @now()")
+      assert {:ok, ["hi ", [{"now", [1, 2]}]], "", _, _, _} = Parser.parse("hi @now(1, 2)")
+      assert {:ok, ["hi ", [{"now", [1, 2]}]], "", _, _, _} = Parser.parse("hi @(now(1, 2))")
+    end
+
+    test "expresssion/1 nested" do
```
JuliaMathias:
```suggestion
    test "expression/1 nested" do
```

test/expression/v2/parser_test.exs
```diff
@@ -0,0 +1,211 @@
...
+    test "expression/1" do
+      assert {:ok, [[{"+", [1, 1]}]], "", _, _, _} = Parser.parse("@(1 + 1)")
+    end
+
+    test "expression/1 with functions" do
+      assert {:ok, [[{"+", [{"+", [{"foo", []}, 1]}, 1]}]], "", _, _, _} =
+               Parser.parse("@(foo() + 1 + 1)")
+    end
+
+    test "operator precendence" do
```
JuliaMathias:
```suggestion
    test "operator precedence" do
```

JuliaMathias (commented):
I've mostly suggested some fixes to typos and added a few comments, but other than that this implementation is looking good to me.

benchmark.exs
smn: Not yet but I'm not too concerned about that, I think understandable code is a good enough reason to take the performance hit.

lib/expression/v2/callbacks.ex
smn: We'll need to copy over the stuff from the existing callbacks and rework them to use the new format. I'm still wanting a way to have each function describe by its own name so we can document them easily using the tools we have instead of having a pile of `callback/3` functions that pattern match.

We'll get to that, right now I'm mostly trying to make sure that the implementation itself is good and then see what we can do to make it easier to document & implement the callbacks.

smn:
thanks for the review @JuliaMathias ! Really appreciate it.

lib/expression/v2/callbacks.ex
smn: This was handled in #102 and has been merged into this PR

PR #101, opened 2023-02-21
Remove decimal dep

smn:
This comes out of #100 , it turns out we can do without the Decimal package and having it adds all sorts of added complexity when we need to do basic arithmatic like > or >= etc. Using floats has Elixir handle that for us directly instead.


smn:
I'll write a test to check that comparisons between floats & ints now work as expected

PR #102, opened 2023-02-28
Feature/revisit parser callbacks

smn:
This builds on the work in #100 and moves the existing callbacks over to the new parser setup.

### DIFF
--- a/lib/expression/v2.ex
+++ b/lib/expression/v2.ex
@@ -105,10 +105,10 @@ defmodule Expression.V2 do
   """
   @spec eval(String.t(), context :: Context.t()) :: [term]
   def eval(expression, context \\ Context.new()) when is_binary(expression) do
-    ast = compile(expression)
-
-    Enum.map(ast, fn
-      part when is_list(part) -> Enum.map(part, &eval_in_context(&1, context)) |> hd()
+    expression
+    |> compile()
+    |> Enum.map(fn
+      part when is_list(part) -> Enum.map(part, &eval_in_context(&1, context)) |> List.first()
       other -> other
     end)
   end
--- a/lib/expression/v2.ex
+++ b/lib/expression/v2.ex
@@ -128,6 +128,14 @@ defmodule Expression.V2 do
 
   def eval_in_context(item, _context), do: item
 
+  @doc """
+  Evaluate an expression and cast all items to strings before joining
+  the full result into a single string value to be returned.
+
+  This calls `eval/2` internally, maps the results with `default_value/2`
+  followed by `stringify/1` and then joins them.
+  """
+  @spec eval_as_string(String.t(), Context.t()) :: String.t()
   def eval_as_string(expression, context \\ Context.new()) do
     eval(expression, context)
     |> Enum.map(&default_value(&1, context))
@@ -141,6 +149,7 @@ defmodule Expression.V2 do
   returned then we can to use the `__value__` value when eval'ing against
   operators or functions.
   """
+  @spec default_value(term) :: term
   def default_value(val, context \\ nil)
   def default_value(%{"__value__" => default_value}, _context), do: default_value
   def default_value(value, _context), do: value
--- a/lib/expression/v2.ex
+++ b/lib/expression/v2.ex
@@ -113,8 +113,7 @@ defmodule Expression.V2 do
   def eval_as_string(expression, context \\ Context.new()) do
     eval(expression, context)
     |> Enum.map(&default_value(&1, context))
-    |> Enum.map(&stringify/1)
-    |> Enum.join("")
+    |> Enum.map_join("", &stringify/1)
   end
 
   @doc """
--- a/lib/expression/v2.ex
+++ b/lib/expression/v2.ex
@@ -60,6 +60,9 @@ defmodule Expression.V2 do
   alias Expression.V2.Context
   alias Expression.V2.Parser
 
+  @doc """
+  Parse a string with expressions into AST for the compile step
+  """
   @spec parse(String.t()) ::
           {:ok, [term]}
           | {:error, reason :: String.t(), bad_parts :: String.t()}
@@ -73,6 +76,11 @@ defmodule Expression.V2 do
     end
   end
 
+  @doc """
+  Parse a string with an expression block into AST for the compile step
+  """
+  @spec parse_block(String.t()) ::
+          {:ok, [term]} | {:error, reason :: String.t(), bad_parts :: String.t()}
   def parse_block(expression_block) do
     case Parser.expression(expression_block) do
       {:ok, ast, "", _, _, _} -> {:ok, ast}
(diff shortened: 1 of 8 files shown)


lib/expression/v2/autodoc.ex
```diff
@@ -0,0 +1,266 @@
+defmodule Expression.V2.Autodoc do
```
smn, line 1: copied from V1 and updated for the new approach, mostly removing a lot of `eval!` and `eval_args!` calls.

lib/expression/v2/callbacks.ex
```diff
@@ -1,52 +1,108 @@
 defmodule Expression.V2.Callbacks do
   @moduledoc """
```
smn, line 2: copied from Expression.Callbacks and updated for the new setup

lib/expression/v2/callbacks/standard.ex
```diff
@@ -0,0 +1,1311 @@
+defmodule Expression.V2.Callbacks.Standard do
```
smn, line 1: copied from Expression.Callbacks.Standard and updated for the new setup, removed a lot of eval! and eval_args! calls

lib/expression/v2.ex
```diff
@@ -73,28 +76,80 @@ defmodule Expression.V2 do
...
+    function.(context)
+  end
 
-    Enum.map(parts, fn
-      part when is_list(part) ->
-        part
-        |> Enum.map(&eval_part(&1, context))
-        |> hd()
+  @doc """
+  Evaluate a string with expressions against a give context
```
JuliaMathias:
```suggestion
  Evaluate a string with expressions against a given context
```

lib/expression/v2/callbacks/standard.ex
```diff
@@ -0,0 +1,1311 @@
...
+         [[_first | _remainder]] <- Regex.scan(regex, String.trim(expression), capture: :all) do
+      # Future match result: first
+      true
+    else
+      _ -> false
+    end
+  end
+
+  @doc """
+  Tests whether `expresssion` contains a phone number.
```
JuliaMathias:
```suggestion
  Tests whether `expression` contains a phone number.
```

JuliaMathias (approved):
Looking good to me! Just suggested a few more typo fixes.

lib/expression/v2/callbacks/standard.ex
```diff
@@ -0,0 +1,1311 @@
+defmodule Expression.V2.Callbacks.Standard do
+  @moduledoc """
+  Callback functions to be used in Expressions.
+
+  This is the same idea as `Expression.Callbacks.Standard` but 
+  it's in a rough shape, mostly to just prove that this all works.
```
santiagocardo, line 6: is this module completely copied already? if so, I think we can modify this moduledoc

lib/expression/v2/context.ex
```diff
@@ -5,14 +5,16 @@ defmodule Expression.V2.Context do
   This will be expanded with support for more attributes that a callback function 
   can access but normal Expression evaluation can not.
   """
-  defstruct vars: %{}, private: %{}, callback_module: Expression.V2.Callbacks
+  defstruct vars: %{}, private: %{}, callback_module: Expression.V2.Callbacks.Standard
```
santiagocardo, line 8: should the `callback_module` be `Expression.V2.Callbacks.Standard` instead of `Expression.V2.Callbacks`?

lib/expression/v2.ex
```diff
@@ -73,28 +76,80 @@ defmodule Expression.V2 do
...
+      {:error, _ast, remaining, _, _, _} -> {:error, "Unable to parse remainder", remaining}
+    end
+  end
+
+  @doc """
+  Evaluate a string with an expression block against a context
+  """
+  @spec eval_block(String.t(), context :: Context.t()) :: term
+  def eval_block(expression_block, context \\ Context.new()) do
+    {:ok, ast} = parse_block(expression_block)
```
santiagocardo: should we handle the **error** case here as well?

lib/expression/v2.ex
```diff
@@ -73,28 +76,80 @@ defmodule Expression.V2 do
...
+  """
+  @spec eval(String.t(), context :: Context.t()) :: [term]
+  def eval(expression, context \\ Context.new()) when is_binary(expression) do
+    ast = compile(expression)
 
-      other ->
-        other
+    Enum.map(ast, fn
+      part when is_list(part) -> Enum.map(part, &eval_in_context(&1, context)) |> hd()
+      other -> other
```
santiagocardo: just a small optional change:

```suggestion
    expression
    |> compile()
    |> Enum.map(ast, fn
      part when is_list(part) -> Enum.map(part, &eval_in_context(&1, context)) |> List.first()
      other -> other
    end)
```

lib/expression/v2.ex
```diff
@@ -73,28 +76,80 @@ defmodule Expression.V2 do
...
+  def eval_in_context(function, context) when is_function(function), do: function.(context)
 
-  def eval_part(atom, context) when is_binary(atom),
+  def eval_in_context(atom, context) when is_binary(atom),
     do: Map.get(context.vars, atom, atom)
 
-  def eval_part(literal, _context), do: literal
+  def eval_in_context(item, _context), do: item
+
+  def eval_as_string(expression, context \\ Context.new()) do
```
santiagocardo, line 139: can we add `doc`and `spec` here?

test/expression/v2/eval_test.exs
```diff
@@ -61,11 +61,7 @@ defmodule Expression.V2.EvalTest do
     end
 
     test "functions with vars" do
-      assert 50 == eval("echo(10) * 5")
-    end
-
-    test "functions vars & properties" do
```
santiagocardo, line 67: something relevant to have removed this test?

santiagocardo (approved):
Looking great! just some minor comments

lib/expression/v2/callbacks/standard.ex
smn: It’s not entirely copied. Had to make some modifications. I’ll update the module doc when we remove the old implementation after this V2 has been tried in production use cases for a bit

test/expression/v2/eval_test.exs
smn: Echo isn’t a standard function and so I reworked these to not use echo. 

The properties on a function is already tested elsewhere by calling .match on the has_all_words function result.

lib/expression/v2/callbacks/standard.ex
santiagocardo: got it!

test/expression/v2/eval_test.exs
santiagocardo: I see! thanks for clarifying

lib/expression/v2.ex
smn: done in cc5760a

lib/expression/v2.ex
smn: done in 69d0c1e

lib/expression/v2.ex
smn: done in e066f3f

lib/expression/v2/context.ex
smn: it is though? Maybe this comment was on an older commit where that still was `.Callbacks`

PR #103, opened 2023-03-01
fix handling of default values in builtins and callbacks

smn:
When we evaluate things in a function or using built ins we should use the __value__ of any complex values used.
Otherwise we end up in a situation where `foo * 5` could be evaluated as `%{"__value__" => 10} * 5` which would fail when instead we just want to do `10 * 5`.

### DIFF
--- a/lib/expression/v2/callbacks.ex
+++ b/lib/expression/v2/callbacks.ex
@@ -53,14 +53,14 @@ defmodule Expression.V2.Callbacks do
         ) :: any
   def callback(module \\ Standard, context, function_name, arguments) do
     case implements(module, function_name, arguments) do
-      {:exact, module, function_name, _arity} ->
+      {:exact, module, function_name} ->
         apply(
           module,
           function_name,
           [context] ++ Enum.map(arguments, &Expression.V2.default_value(&1, context))
         )
 
-      {:vargs, module, function_name, _arity} ->
+      {:vargs, module, function_name} ->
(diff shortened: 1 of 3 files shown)


smn:
This is a change coming out of doing a trial run in https://github.com/turnhub/flow_runner/pull/99

lib/expression/v2/callbacks.ex
```diff
@@ -54,10 +54,17 @@ defmodule Expression.V2.Callbacks do
   def callback(module \\ Standard, context, function_name, arguments) do
     case implements(module, function_name, arguments) do
       {:exact, module, function_name, _arity} ->
```
santiagocardo: just curiousity: is this `implements/3` function used anywhere else? I see we are not using `arity` at all

santiagocardo (approved):
LGTM!

lib/expression/v2/callbacks.ex
smn: No, afaik it's only used here? I copied it from your original implementation :)

santiagocardo: I think it has always returned the `arity` but we are not using it anywhere. Should we just refactor it?

smn: yeah feel free to do so in a separate PR ?

smn: actually I still have some stuff to do here, I'll include it here.

PR #104, opened 2023-03-02
fix problem with nested properties & attributes

smn:
Updates coming out of running this against 2 repos that use expressions.

### DIFF
--- a/lib/expression/v2.ex
+++ b/lib/expression/v2.ex
@@ -76,6 +76,19 @@ defmodule Expression.V2 do
     end
   end
 
+  @spec escape(String.t()) :: String.t()
+  def escape(expression) when is_binary(expression) do
+    String.replace(expression, ~r/@([a-z]+)(\(|\.)/i, "@@\\g{1}\\g{2}")
+  end
+
+  @doc """
+  This function is referenced by `Expression.V2.Compile` to 
+  make access to values in Maps or Lists easier
+  """
+  @spec read_attribute(map | list, binary | integer) :: term
+  def read_attribute(map, item) when is_map(map), do: Map.get(map, item)
+  def read_attribute(list, index) when is_list(list), do: Enum.at(list, index)
+
   @doc """
   Parse a string with an expression block into AST for the compile step
   """
@@ -84,6 +97,7 @@ defmodule Expression.V2 do
   def parse_block(expression_block) do
     case Parser.expression(expression_block) do
       {:ok, ast, "", _, _, _} -> {:ok, ast}
+      {:ok, _ast, _remainder, _, _, _} -> {:error, "Unable to parse", expression_block}
       {:error, _ast, remaining, _, _, _} -> {:error, "Unable to parse remainder", remaining}
     end
   end
--- a/lib/expression/v2/callbacks.ex
+++ b/lib/expression/v2/callbacks.ex
@@ -109,7 +109,7 @@ defmodule Expression.V2.Callbacks do
       def callback(module \\ __MODULE__, context, function_name, args)
 
       def callback(module, context, built_in, args)
-          when built_in in ["*", "+", "-", ">", ">=", "<", "<=", "/", "^", "=="],
+          when built_in in ["*", "+", "-", "<>", ">", ">=", "<", "<=", "/", "^", "=="],
           do:
             apply(
               Kernel,
--- a/lib/expression/v2/compile.ex
+++ b/lib/expression/v2/compile.ex
@@ -62,11 +62,24 @@ defmodule Expression.V2.Compile do
   """
   @spec wrap_in_context(Macro.t()) :: Macro.t()
   def wrap_in_context(quoted) do
+    # Check to see if the generated AST makes a reference to the context.
+    # If that is the case then generate an AST that makes it available.
+    # If there are no references to the context then prefix the variable
+    # with an underscore to keep the compiler happy and not emit warnings
+    # at runtime
+    {quoted, uses_context?} =
+      Macro.prewalk(quoted, false, fn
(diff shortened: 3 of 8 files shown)


lib/expression/v2.ex
```diff
@@ -76,6 +76,19 @@ defmodule Expression.V2 do
     end
   end
 
+  @spec escape(String.t()) :: String.t()
+  def escape(expression) when is_binary(expression) do
+    String.replace(expression, ~r/@([a-z]+)(\(|\.)/i, "@@\\g{1}\\g{2}")
+  end
```
smn, line 82: This is just copied from the original Expressions module. Copying here rather than defdelegating as it's tiny and it'll make search & replacing `Expressions.V2` easier when we want to start doing that.

lib/expression/v2.ex
```diff
@@ -84,6 +97,7 @@ defmodule Expression.V2 do
   def parse_block(expression_block) do
     case Parser.expression(expression_block) do
       {:ok, ast, "", _, _, _} -> {:ok, ast}
+      {:ok, _ast, _remainder, _, _, _} -> {:error, "Unable to parse", expression_block}
```
smn, line 100: When a block is invalid, the parser will return `{:ok, }` up until where it got to. For us this means the whole thing is still garbage and we should treat it as such.

lib/expression/v2/callbacks.ex
```diff
@@ -109,7 +109,7 @@ defmodule Expression.V2.Callbacks do
       def callback(module \\ __MODULE__, context, function_name, args)
 
       def callback(module, context, built_in, args)
-          when built_in in ["*", "+", "-", ">", ">=", "<", "<=", "/", "^", "=="],
+          when built_in in ["*", "+", "-", "<>", ">", ">=", "<", "<=", "/", "^", "=="],
```
smn, line 112: Some of our tests use `<>` to join two strings together. I'm not sure if that's standard (I doubt it) but putting it here for backwards compatibility.

lib/expression/v2/compile.ex
```diff
@@ -107,39 +120,17 @@ defmodule Expression.V2.Compile do
...
-
-  defp quoted({"__attribute__", [a, b]}) when is_integer(b) do
-    # When the attribute we're trying to read is an integer then we're
-    # trying to read an index off of a list.
-    # `foo[1]` becomes `Enum.at(foo, 1)`
-    {{:., [], [Enum, :at]}, [],
-     [
-       quoted(a),
-       quoted(b)
-     ]}
```
smn, line 126: `foo["bar"]` for properties ended up being a lot more concise and readable than a lot of nested Map.get calls.

lib/expression/v2/compile.ex
```diff
@@ -107,39 +120,17 @@ defmodule Expression.V2.Compile do
...
-          a
-        ]},
-       quoted(b)
-     ]}
+    # Since Map keys in Expressions can either be integers or strings
+    # we use the helper in Expression.V2.read_attribute to read
+    # the correct value using Elixir function guards in compiled
+    # code rather than attempting to generate the AST for that here.
+    {{:., [], [{:__aliases__, [alias: false], [:Expression, :V2]}, :read_attribute]}, [],
+     [quoted(a), quoted(b)]}
```
smn, line 133: See the `read_attribute/2` added in Expression.V2

lib/expression/v2/compile.ex
```diff
@@ -178,11 +169,7 @@ defmodule Expression.V2.Compile do
   end
 
   defp quoted(atom) when is_binary(atom) do
-    {{:., [], [{:__aliases__, [alias: false], [:Map]}, :get]}, [],
-     [
-       {{:., [], [{:context, [], nil}, :vars]}, [no_parens: true], []},
-       atom
-     ]}
+    {{:., [], [Access, :get]}, [], [context_dot_vars(), atom]}
```
smn, line 172: again `context.vars["atom"]` ended up being more readable than lots of nested `Map.get` calls

test/expression/v2/eval_test.exs
```diff
@@ -40,6 +42,17 @@ defmodule Expression.V2.EvalTest do
...
 
+    test "attribute and property combination" do
+      assert "hi" ==
+               eval(
+                 "after_hours.rows[random_row].label.bar",
+                 %{
+                   "after_hours" => %{"rows" => [%{"label" => %{"bar" => "hi"}}]},
+                   "random_row" => 0
+                 }
+               )
```
smn, line 54: test for nested attributes & properties

test/expression/v2/eval_test.exs
```diff
@@ -117,5 +130,21 @@ defmodule Expression.V2.EvalTest do
...
+               ["three", "Button Three"],
+               ["four", "Button Four"],
+               ["five", "Button Five"]
+             ] ==
+               eval(
+                 ~S|map(options, &([&1, concatenate('Button ', proper(&1))]))|,
+                 %{
+                   "options" => ["one", "two", "three", "four", "five"]
+                 }
+               )
```
smn, line 148: test for handling single quoted strings

test/expression/v2_test.exs
```diff
@@ -3,16 +3,22 @@ defmodule Expression.V2Test do
...
+                 [context.vars["foo"], context.vars["bar"]]
                ])
              end
              """) == V2.debug("foo + bar")
     end
+
+    test "code gen not using context" do
+      assert String.trim("""
+             fn _context -> ["one", "two"] end
+             """) == V2.debug(~S|["one", "two"]|)
```
smn, line 21: test for not emitting warnings when context isn't used in an expression

PR #105, opened 2023-03-02
Fix regex issue in single digit integers

smn:


lib/expression/v2/callbacks/standard.ex
```diff
@@ -1146,6 +1148,9 @@ defmodule Expression.V2.Callbacks.Standard do
   @expression_doc expression: "has_number_lte(\"the number is 42.0\", \"40\")", result: false
   @expression_doc expression: "has_number_lte(\"the number is 40\", \"foo\")", result: false
   @expression_doc expression: "has_number_lte(\"four hundred\", \"foo\")", result: false
+  @expression_doc expression: "has_number_lte(response, 5)",
```
smn, line 1151: there may still need to come a fix in for expressions in expressions. Notice that for V1 we supported `has_number_lte("@response", 5)` whereas in V2 we do not yet. I'll get a different PR in when that becomes clearer.

PR #110, opened 2023-03-20
v1 -> v2 compatibility module

smn:
Introduces an `Expression.V2.Compat` module which is a drop in replacement for `Expression`, exposing the same functions with the same arity and returning the same result.

This is useful for systems wanting to migrate from the existing implementation to the new implementation.


lib/expression/callbacks/standard.ex
```diff
@@ -121,8 +121,10 @@ defmodule Expression.Callbacks.Standard do
                   result: ~D[2022-01-01]
   def datevalue(ctx, date, format) do
     [date, format] = eval!([date, format], ctx)
-    date = DateHelpers.extract_dateish(date)
-    %{"__value__" => Timex.format!(date, format, :strftime), "date" => date}
+
+    if date = DateHelpers.extract_dateish(date) do
```
smn, line 125: not entirely related bugfix, it turns out that `date` can be `nil` and that'll cause this to crash otherwise.

lib/expression/v2.ex
```diff
@@ -82,12 +82,17 @@ defmodule Expression.V2 do
...
   """
   @spec read_attribute(map | list, binary | integer) :: term
   def read_attribute(map, item) when is_map(map), do: Map.get(map, item)
-  def read_attribute(list, index) when is_list(list), do: Enum.at(list, index)
+
+  def read_attribute(list, index) when is_list(list) and is_integer(index),
+    do: Enum.at(list, index)
+
+  def read_attribute(list, range) when is_list(list) and is_struct(range, Range),
+    do: Enum.slice(list, range)
```
smn, line 95: `foo[0...3]` is a syntax that's supported in V1, needed to introduce it in to V2.

lib/expression/v2.ex
```diff
@@ -109,38 +114,47 @@ defmodule Expression.V2 do
           term | {:error, reason :: String.t(), bad_parts :: String.t()}
   def eval_block(expression_block, context \\ Context.new()) do
     with {:ok, ast} <- parse_block(expression_block) do
-      function = Compile.compile(ast)
-      function.(context)
+      hd(eval_block_ast(ast, context))
```
smn, line 124: the changes below are me reworking `eval` to use `eval_block` internally and to end up with some smaller reusable functions that can operate on AST.

lib/date_helpers.ex
```diff
@@ -155,6 +155,7 @@ defmodule Expression.DateHelpers do
   def extract_dateish(date_time) when is_struct(date_time, DateTime), do: date_time
   def extract_dateish(date) when is_struct(date, Date), do: date
   def extract_dateish(nil), do: nil
+  def extract_dateish({:not_found, _}), do: nil
```
smn, line 158: same as below, a date passed could be `nil` or not found.

nathanbegbie (approved):
Rubber Stamp

PR #116, opened 2023-04-14
contextvars was over-engineered, simplify things

smn:
The ContextVars existed to allow us to keep track of the variable reference when it doesn't point at anything so we could still render `please contact us via support@some.domain` when `some` doesn't exist in the context.

It was a very elaborate solution to something that only really needed to be solved in the `eval_as_string` function. Since we could use the original AST there to keep track of what the original reference variables were, we could just solve it there rather than the elaborate Access behaviour implementation in ContextVars.

This also removes the need for the `truthy` helper and allows us to fallback to built-in functions for things like `+` etc.


lib/expression/v2.ex
```diff
@@ -174,11 +167,36 @@ defmodule Expression.V2 do
...
+    |> Enum.zip(ast)
+    |> Enum.map_join("", fn
+      {nil, [{"__property__", _parts} = property]} ->
+        "@" <> unwrap_property(property)
+
+      {value, _ast} ->
+        value
+        |> default_value(context)
+        |> stringify()
+    end)
```
smn, line 183: this is what the change is about, the existing tests cover this and confirm that it's still working as expected.

santiagocardo (approved):
Nice refactor! Much cleaner 🍰

PR #119, opened 2023-05-04
Truncate datetime microseconds to precision 6

fedme:
It seems like Elixir's `DateTime` only supports microseconds up to precision 6, which causes our parser to throw exceptions for datetime strings containing microseconds with precision greater than 6.

See for example this code from Elixir's source code: https://github.com/elixir-lang/elixir/blob/6f96693b355a9b670f2630fd8e6217b69e325c5a/lib/elixir/lib/calendar/iso.ex#L242

Also notice DateTime truncating microseconds to precision 6 when parsing an ISO8601 string with precision 7 (notice the "2" getting dropped):
```
iex(2)> DateTime.from_iso8601("2022-11-10T13:40:05.9213782Z")
{:ok, ~U[2022-11-10 13:40:05.921378Z], 0}
```

This PR updates `DateHelpers.to_datetime/1` so that it drops microseconds greater than precision 6.

### DIFF
--- a/lib/date_helpers.ex
+++ b/lib/date_helpers.ex
@@ -109,13 +109,17 @@ defmodule Expression.DateHelpers do
 
           {microseconds, values} = Keyword.pop(values, :microsecond, 0)
 
-          microsecond_entry =
-            {microseconds,
-             microseconds
-             |> to_string()
-             |> String.length()}
+          # Elixir's DateTime only supports microseconds up to precision 6
+          microseconds_string = to_string(microseconds)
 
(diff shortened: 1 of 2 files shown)


nathanbegbie (approved):
Nice work @fedme

fedme:

fedme:
recheck

nathanbegbie:
@fedme apparently I don't have write access to this repo?

Maybe @shaunsephton does?

Screenshot 2023-05-04 at 17 04 20

nathanbegbie (approved):
Approving again. Because CLA bot?

PR #120, opened 2023-05-16
Give datetime comparisons some more breathing room on slower CIs and use literals with builtin functions

smn:
1. The Compat module is useful but also too strict on slower machines/CI because it'll fail on microsecond differences for 
expressions involving dynamically generated timestamps. 
2. When we're doing `1 * 1` and both sides of the operator are integers or strings, there's no need to call `default_value()` on them as that won't have any effect. Side step that function call as an optimisation, mostly to generate more readable debug output and save the CPU some function invocation work during eval.


fedme (approved):
lgtm

PR #126, opened 2023-05-26
[sc-20833] fix(callbacks): support for numbers in has_all_words/2

santiagocardo:
Add support for numbers in `has_all_words/2` from the Standard callbacks module. Also, add a refactor to evaluate `:function` expressions in a lazy fashion.

### DIFF
--- a/lib/expression/callbacks/standard.ex
+++ b/lib/expression/callbacks/standard.ex
@@ -905,7 +905,7 @@ defmodule Expression.Callbacks.Standard do
 
     results =
       patterns
-      |> Enum.map(&Regex.run(&1, haystack))
+      |> Enum.map(&Regex.run(&1, to_string(haystack)))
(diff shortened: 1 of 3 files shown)


lib/expression/callbacks/standard.ex
```diff
@@ -905,7 +905,7 @@ defmodule Expression.Callbacks.Standard do
 
     results =
       patterns
-      |> Enum.map(&Regex.run(&1, haystack))
+      |> Enum.map(&Regex.run(&1, to_string(haystack)))
```
santiagocardo, line 908: this is the actual fix

lib/expression/eval.ex
santiagocardo, line 1: this is the refactor to evaluate `function` expressions in a lazy fashion.

smn (approved):

PR #151, opened 2023-11-06
handle empty list evaluation

smn:
We're not handling `[]` properly, it breaks because the reducer assumes lists will always have members.


santiagocardo (approved):
LGTM!

PR #155, opened 2023-11-15
feat: add append & delete functions

santiagocardo:
Add `append` & `delete` functions to standard expression callbacks.

### DIFF
--- a/lib/expression/callbacks/standard.ex
+++ b/lib/expression/callbacks/standard.ex
@@ -1449,13 +1449,8 @@ defmodule Expression.Callbacks.Standard do
   @doc """
   Deletes an element from a map by the given key.
   """
-  @expression_doc expression: "delete(contact, \"gender\")",
-                  context: %{
-                    "contact" => %{
-                      "gender" => "?",
-                      "age" => 32
-                    }
-                  },
+  @expression_doc expression: "delete(patient, \"gender\")",
+                  context: %{"patient" => %{"gender" => "?", "age" => 32}},
                   result: %{"age" => 32}
   def delete(ctx, map, key) do
     [map, key] = eval_args!([map, key], ctx)
--- a/lib/expression/callbacks/standard.ex
+++ b/lib/expression/callbacks/standard.ex
@@ -1436,14 +1436,17 @@ defmodule Expression.Callbacks.Standard do
   end
 
   @doc """
-  Appends items from one list to another list.
+  Appends an item or a list of items to a given list.
   """
+  @expression_doc expression: "append([\"A\", \"B\"], \"C\")",
+                  result: ["A", "B", "C"]
   @expression_doc expression: "append([\"A\", \"B\"], [\"C\", \"B\"])",
                   result: ["A", "B", "C", "B"]
-  def append(ctx, first_list, second_list) do
-    [first_list, second_list] = eval_args!([first_list, second_list], ctx)
+  def append(ctx, list, payload) do
+    [list, payload] = eval_args!([list, payload], ctx)
+    enumerable = if is_list(payload), do: payload, else: [payload]
 
-    Enum.concat(first_list, second_list)
+    Enum.concat(list, enumerable)
   end
 
   @doc """
--- a/lib/expression/callbacks/standard.ex
+++ b/lib/expression/callbacks/standard.ex
@@ -1434,4 +1434,32 @@ defmodule Expression.Callbacks.Standard do
 
     rem(integer1, integer2)
   end
+
+  @doc """
+  Appends items from one list to another list.
+  """
+  @expression_doc expression: "append([\"A\", \"B\"], [\"C\", \"B\"])",
+                  result: ["A", "B", "C", "B"]
+  def append(ctx, first_list, second_list) do
+    [first_list, second_list] = eval_args!([first_list, second_list], ctx)
+
+    Enum.concat(first_list, second_list)
+  end
+
+  @doc """
+  Deletes an element from a map by the given key.
+  """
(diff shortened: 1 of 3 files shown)


lib/expression/callbacks/standard.ex
```diff
@@ -1434,4 +1434,32 @@ defmodule Expression.Callbacks.Standard do
 
     rem(integer1, integer2)
   end
+
+  @doc """
+  Appends items from one list to another list.
+  """
+  @expression_doc expression: "append([\"A\", \"B\"], [\"C\", \"B\"])",
+                  result: ["A", "B", "C", "B"]
+  def append(ctx, first_list, second_list) do
```
fedme: Usually a function called `append` will add one element to a list. If this takes 2 lists should we call it `concat` rather?

lib/expression/callbacks/standard.ex
```diff
@@ -1434,4 +1434,32 @@ defmodule Expression.Callbacks.Standard do
...
+
+    Enum.concat(first_list, second_list)
+  end
+
+  @doc """
+  Deletes an element from a map by the given key.
+  """
+  @expression_doc expression: "delete(contact, \"gender\")",
+                  context: %{
+                    "contact" => %{
```
fedme: Can we use a different example? Using `contact` here seems to suggest the you can remove fields forma contact profile with this function

test/expression_test.exs
```diff
@@ -49,6 +49,14 @@ defmodule ExpressionTest do
       assert nil == Expression.evaluate!("@foo[100]", %{"foo" => ["baz", "bar"]})
     end
 
+    test "append lists" do
```
fedme: Can we add corner cases to the test suite? Like empty lists, nils, etc.

lib/expression/callbacks/standard.ex
santiagocardo: yeah, that's true. But there is also a function called `concatenate` that concatenates strings 😅

lib/expression/callbacks/standard.ex
santiagocardo: good point!

test/expression_test.exs
santiagocardo: I can do it but in the current **expression** implementation, it will actually throw an error.

lib/expression/callbacks/standard.ex
fedme: I see, very unfortunate! :D
Can we maybe have this append function accept both a list or a single element as the second argument?

santiagocardo: what about `merge`? maybe we can add support for maps too.

santiagocardo: or maybe just use `concat` as you suggested but also as we can see from [TextIt expressions](https://textit.in/mr/docs/expressions.html#function:concat).

test/expression_test.exs
smn: I think https://github.com/turnhub/expression/pull/151 fixes this issue -- would appreciate a review there :)

lib/expression/callbacks/standard.ex
smn: either we're going to have to be very disciplined about naming conventions (follow Redis' command examples) or we need to start namespacing into modules.

test/expression_test.exs
santiagocardo: Approved @smn!
But in the current implementation, [we are ok throwing errors](https://github.com/turnhub/expression/blob/develop/test/expression_test.exs#L332) when manipulating incompatible types such as `nil`, `map`, and `string`. That is something we can revisit later in another PR or lib version.

lib/expression/callbacks/standard.ex
santiagocardo: what do you think it's better?
- `append` supporting both, a single element and a list of elements
- `merge` supporting both, lists and maps
- `concat` just for lists

I need to wrap up this because some people are waiting for this right now.

fedme: if `concatenate` already exists in the expressions we support, should we maybe make it work for lists too in addition to strings?

santiagocardo: yes, that's also a good option. I like it!

santiagocardo: @smn what do you think about this? ⬆️

test/expression_test.exs
smn: yeah I think we're ok with that. We need to form some good opinions on how to handle errors in expressions anyway, both when they're used as a standalone block `foo > bar` and when they're using in a string `Hello @if(foo > bar, "Anonymous", "Member")`

lib/expression/callbacks/standard.ex
smn: I like that idea but keep in mind that concatenate() as it's defined in the spec takes any amount of string arguments and joins them. We'd need to then do the same for lists, so it would be like a combination of Enum.concat and Enum.reduce

smn:
@santiagocardo happy for you to wrap this up as best as you see fit!

smn (approved):
I think @fedme 's review on this is great! Let's address those things as best we can and then merge.

lib/expression/callbacks/standard.ex
santiagocardo: Ok, I ended up using `append()` which is what two clients were asking for but also to simplify things 👍🏼

test/expression_test.exs
santiagocardo: Agree @smn. We would likely change that behaviour in the next `expression` version.

PR #156, opened 2023-11-21
Extend has_phrase function to support nils and numbers

fedme:
The `has_phrase` callback currently throws an exception if any of the arguments are null or integers.

This PR extends `has_phrase` to support null values and integer values too.


santiagocardo (approved):
LGTM!

Can we also do it for `has_only_phrase`?

fedme:
> LGTM!
> 
> Can we also do it for `has_only_phrase`?

Will do @santiagocardo!

PR #159, opened 2023-12-05
fix issue for reading keys and attributes from non-iterables

smn:
We crash on something like `foo.bar` where `foo` isn't a map, we should just return a `nil` there.
The same for `foo[a]` when `foo` isn't a list or a map


fedme (approved):
LGTM

PR #160, opened 2023-12-07
Fix concatenation with default values

smn:
`datevalue(now(), "%Y-%m-%d")` returns a complex value:

```elixir
iex(1)> Expression.evaluate_block("datevalue(now(), \"%Y-%m-%d\")")
{:ok, %{"__value__" => "2023-12-07", "date" => ~U[2023-12-07 09:50:34.053039Z]}}
```

If we use that in `concatenate(datevalue(thing.date, "%Y-%m-%d"), "T", thing.time)` to construct a new datetime then it fails because it tries to join the complex value rather than the default value.

This fixes that by retrieving the default value before attempting to join things.


fedme (approved):
LGTM

PR #161, opened 2023-12-07
use datetime for datevalue function

smn:
When giving this function a datetime value we drop the `%H:%m:%s` precision because we're treating it as a date.
This has it work with datetime internally and assumes `00:00:00` when no time is present.


fedme (approved):
LGTM

PR #162, opened 2023-12-07
return matched values on has_* functions

smn:
Knowing that we found something is great, having access to it is greater.

when we do `has_number("some multiple of 8")` this will return:

```elixir
%{"__value__" => true, "number" => 8.0}
```

and so it can be used as a test function as its default value is `true` but one can also read the matched number with `.number`

### DIFF
--- a/lib/expression/callbacks/standard.ex
+++ b/lib/expression/callbacks/standard.ex
@@ -1378,7 +1378,7 @@ defmodule Expression.Callbacks.Standard do
     parse_phone_number(letters_removed, country_code)
   end
 
-  def parse_phone_number(string, country_code) do
+  defp parse_phone_number(string, country_code) do
     pn =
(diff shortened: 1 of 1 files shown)


lib/expression/callbacks/standard.ex
```diff
@@ -1332,33 +1356,36 @@ defmodule Expression.Callbacks.Standard do
...
 
-    case ExPhoneNumber.parse(letters_removed, country_code) do
-      # Future match result: ExPhoneNumber.format(pn, :es164)
-      {:ok, _pn} -> true
-      _ -> false
-    end
+    parse_phone_number(letters_removed, country_code)
+  end
+
+  def parse_phone_number(string, country_code) do
```
rudigiesler: Do we maybe want this to be a private `defp` function?

rudigiesler (approved):
Looks good!

PR #163, opened 2023-12-07
add parse_datevalue()

smn:
Allows us to do `parse_datevalue("2016-02-29T22:25:00-00:00", "%FT%T%:z")` and get back `~U[2016-02-29 22:25:00Z]
`

### DIFF
--- a/lib/expression/callbacks/standard.ex
+++ b/lib/expression/callbacks/standard.ex
@@ -145,7 +145,11 @@ defmodule Expression.Callbacks.Standard do
     }
   end
 
-  @expression_doc doc: "Parse a date value and return a DateTime",
+  @doc """
+  Parse random dates and times with `strftime` patterns and return a DateTime value
+  when it matches.
+  """
+  @expression_doc doc: "Parse a date value using strftime formatting and return a DateTime",
                   expression: "parse_datevalue(\"2016-02-29T22:25:00-00:00\", \"%FT%T%:z\")",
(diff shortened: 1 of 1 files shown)


lib/expression/callbacks/standard.ex
```diff
@@ -145,6 +145,28 @@ defmodule Expression.Callbacks.Standard do
...
   end
 
+  @doc """
+  Parse random dates and times with `strftime` patterns and return a DateTime value
+  when it matches.
+  """
+  @expression_doc doc: "Parse a date value using strftime formatting and return a DateTime",
+                  expression: "parse_datevalue(\"2016-02-29T22:25:00-00:00\", \"%FT%T%:z\")",
+                  result: DateTime.new!(~D[2016-02-29], ~T[22:25:00])
+  @expression_doc doc: "Attempt to parse a date value and return nil when failing",
```
fedme, line 155: are these doctested? If not could we add tests?

fedme (approved):
lgtm

lib/expression/callbacks/standard.ex
smn: these @expression_docs are compiled by the macro to become doctests and documentation at the same time

fedme: really impressive, so cool!

smn: https://hexdocs.pm/expression/Expression.Callbacks.Standard.html#parse_datevalue/3 are the docs + doctests

PR #168, opened 2024-01-11
Fix compilation warnings for 1.16

smn:


lib/expression/callbacks/standard.ex
```diff
@@ -856,13 +856,13 @@ defmodule Expression.Callbacks.Standard do
       stop > 0 ->
         binary
         |> String.split(@punctuation_pattern)
-        |> Enum.slice((start - 1)..(stop - 2))
+        |> Enum.slice((start - 1)..(stop - 2)//1)
```
fedme, line 859: what does this do?

smn: It slices backwards starting at the end. This used to assume a step of 1 when the slice is negative, elixir 1.16 wants the step listed explicitly.

fedme (approved):

PR #170, opened 2024-01-16
do not break validation on built-in operators

smn:
Now that we're doing more complete validation in areas of Turn this issue has surfaced.


santiagocardo (approved):
Nice to see this!

PR #171, opened 2024-01-16
Accept pre-parsed expressions in functions that evaluate expressions

fedme:
Accept pre-parsed expressions (Expressions ASTs instead of strings) in functions that evaluate expressions.

This allows the library user to cache pre-parsed expressions when evaluating them rather than have this library parse the expression AST from an expression string at every evaluation.


lib/expression.ex
```diff
@@ -83,9 +83,11 @@ defmodule Expression do
     String.replace(expression, ~r/@([a-z]+)(\(|\.)?/i, "@@\\g{1}\\g{2}")
   end
 
-  @spec parse!(String.t() | Number.t()) :: Keyword.t()
+  @spec parse!(String.t() | Number.t() | list()) :: Keyword.t()
   def parse!(expression) when is_number(expression), do: to_string(expression) |> parse!()
 
+  def parse!(expression_ast) when is_list(expression_ast), do: expression_ast
```
fedme, line 89: Is there a better way to match Expressions ASTs?

smn: right now there isn't :/

test/expression_test.exs
```diff
@@ -413,4 +424,18 @@ defmodule ExpressionTest do
...
+
+  test "parse!/1 skips expensive parsing is the expression is already an AST struct" do
+    # If the expression is a provided as a string then it gets parsed to an AST
+    assert [expression: [attribute: [atom: "contact", atom: "age"]], text: " + 2"] ==
+             Expression.parse!("@contact.age + 2")
+
+    # If the expression is a already provided as a parsed AST then it is simply
+    # returned as is
+    assert [expression: [attribute: [atom: "contact", atom: "age"]], text: " + 2"] ==
+             Expression.parse!(
```
smn: supplying AST (parsed structure) to a function called `parse!` seems a bit odd to be honest, what problem are we solving here?

lib/expression.ex
santiagocardo: Perhaps you can write a `defguard` function to make sure the given payload contains an `:expression` attribute and the expression payload is valid.

santiagocardo: Otherwise, it can be possible to just give a random list such as `["hola", "mundo"]` which is not valid.

test/expression_test.exs
fedme: This is just because this function is called from inside every `evaluate_*` function, so it was easier to add this check here than adding it to every `evaluate_*` function, but I could also do that. I agree that it feels odd.

As per the PR description the goal here is to allow the library user to cache/store AST representations of commonly evaluated expressions rather than forcing the library user to supply the expression as a string (and waste time parsing it again) at every evaluation.

lib/expression.ex
fedme: Good idea @santiagocardo! Will do tomorrow!

smn:
Closing for now due to inactivity, sorry @fedme !

PR #172, opened 2024-01-30
Support non-string values in has_beginning

fedme:
Ensure that `has_beginning` does not break on non-string values (such as numeric values).
The same was done for `has_phrase` in https://github.com/turnhub/expression/pull/156


Djack1010 (approved):
LGTM :+1:

PR #174, opened 2024-02-01
Fix Autodoc multiline code snippet generation

fedme:
Fixes Autodoc multiline code snippet generation by using multiline code markdown syntax (```) when needed.


smn (approved):
Thank you!

PR #177, opened 2024-03-03
[feature] re-introduce v1 <-> v2 compatibility layer

smn:
We have a V2 ready that addresses a number of problems with V1 (lazy evaluations, some operator precedence bugs etc). It'll likely also introduce some backwards incompatible changes that need to be surfaced. Having a compatibility layer allows V2 to shadow V1 in production environments to get a line of sight of what incompatibilities surface.


lib/expression/v2/compat.ex
```diff
@@ -115,62 +114,56 @@ defmodule Expression.V2.Compat do
...
+        return_or_log(expression, context, v2_resp, v2_resp)
+
+      # Hack for handling `@if` expressions, in V2 these aren't evaluated.
+      # See the note for this in `eval_compat_test.exs`.
+      String.contains?(String.downcase(expression), ["@if", "@left"]) ->
+        return_or_log(expression, context, v2_resp, v2_resp)
+
+      true ->
+        return_or_log(expression, context, v1_resp, v2_resp)
+    end
```
santiagocardo, line 165: I think this can be simplified as we are not doing anything with the different conditions
```suggestion
    return_or_log(expression, context, v1_resp, v2_resp)
```

santiagocardo (approved):
Wow, there are a lot of things I don't get from that `Expression.V2.Compat` module but I'm approving it since most of the changes are putting things back in place.

lib/expression/v2/compat.ex
smn: the difference is we're defaulting to whatever v2 returns here rather than `v1_resp` or `v2_resp`. The difference is subtle but important.

smn:
> Wow, there are a lot of things I don't get from that `Expression.V2.Compat` module but I'm approving it since most of the changes are putting things back in place.

Yeah, it's not the greatest bit of code. I was hoping that it would be a temporary thing but it's been in there for quite a while. I'm hoping that standardising the `Expression.Behaviour` would give us a better interface to create these kinds of implementations where V2 shadows V1 and logs any differences found.

lib/expression/v2/compat.ex
santiagocardo: Ahh got it! missed that. Thanks @smn

smn:
@santiagocardo this has all outstanding PRs now merged in. I'm going to open a PR on Turn to see how badly this breaks our existing tests there by pointing mix.exs to github with this branch.

smn:
closing this as there's no further progress to be made here

PR #185, opened 2024-03-06
create v1 namespace

smn:
This builds on #177 and draws more explicit lines for the V1 vs V2 modules.


santiagocardo (approved):
Thanks! This is a lot more cleaner 🙌🏼

PR #186, opened 2024-03-06
refactor base module

smn:
This builds on #185 and creates a simple base module should we want to start looking at swapping out one for another. This moves V1 things into the `expression/v1` directory and updates module references.


santiagocardo (approved):
Great!

PR #188, opened 2024-03-12
Allow comparing date/datetimes to ISO8601 date/datetime strings

fedme:
This PR adds overloads for the `:=, :==, :!=, :<, :<=, :>, :>=` boolean operators to allows users to compare `Date` or `DateTime` values with ISO8601 string representations of dates and datetimes.

Adding support for this kind of comparisons makes it much easier to create Triggers/Conditionals in Turn that compare a Contact's datetime field to a specific datetime value (expressed as an ISO8601 string).

### DIFF
--- a/lib/expression/eval.ex
+++ b/lib/expression/eval.ex
@@ -177,7 +177,7 @@ defmodule Expression.Eval do
   def op(:==, a, b) when is_struct(a, DateTime) and is_struct(b, DateTime),
     do: DateTime.compare(a, b) == :eq
 
-  def op(:=, a, b) when is_struct(a, Date) and is_struct(b, Date),
+  def op(:=, a, b) when is_struct(a, DateTime) and is_struct(b, DateTime),
     do: Date.compare(a, b) == :eq
 
   def op(:>, a, b) when is_struct(a, Date) and is_struct(b, Date),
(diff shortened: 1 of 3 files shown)


lib/expression/eval.ex
```diff
@@ -177,7 +177,7 @@ defmodule Expression.Eval do
   def op(:==, a, b) when is_struct(a, DateTime) and is_struct(b, DateTime),
     do: DateTime.compare(a, b) == :eq
 
-  def op(:=, a, b) when is_struct(a, Date) and is_struct(b, Date),
+  def op(:=, a, b) when is_struct(a, DateTime) and is_struct(b, DateTime),
```
fedme, line 182: I think this was a typo, as this function clause was appearing twice.
I think this one was supposed to be for `DateTime`, not for `Date`: the one for `Date` already exists down below.

fedme: @smn could you confirm?

smn: yeah, most probably

fedme:
@smn @santiagocardo any ideas on how to fix the failing dialyzer check in CI? It passes locally. It's saying something about an invalid PLT? Maybe an invalid cache?

lib/expression/eval.ex
```diff
@@ -198,6 +198,33 @@ defmodule Expression.Eval do
...
+             is_binary(b) do
+    comparison_result =
+      if is_struct(a, Date) do
+        Date.compare(a, Expression.DateHelpers.extract_dateish(b))
+      else
+        DateTime.compare(a, Expression.DateHelpers.extract_datetimeish(b))
+      end
+
+    case {operator, comparison_result} do
+      {:=, :eq} -> true
```
santiagocardo, line 225: are we intentionally supporting the _match_ operator as an _equal_ operator?

fedme: yes, this was already the case in the existing code. The FLOIP standard supports both `=` and `==` (or at least RapidPro does).

lib/expression/eval.ex
```diff
@@ -198,6 +198,33 @@ defmodule Expression.Eval do
   def op(:=, a, b) when is_struct(a, Date) and is_struct(b, Date),
     do: Date.compare(a, b) == :eq
 
+  # credo:disable-for-next-line Credo.Check.Refactor.CyclomaticComplexity
+  def op(operator, a, b)
+      when operator in [:=, :==, :!=, :<, :<=, :>, :>=] and
+             (is_struct(a, Date) or is_struct(a, DateTime)) and
+             is_binary(b) do
```
smn: @fedme are we handling a scenario where `a` is a binary and `b` is a struct? Do we need to?

smn (approved):
Happy with this change, just curious about when `a` is a binary and `b` is a struct and if we should handle that. I can see how that could be expected from an expression authoring perspective if we're already supporting the scenario we're testing against in this PR.

If we don't, we should document this behaviour well.

lib/expression/eval.ex
fedme: In this PR I am only handling the case where the 1st argument is a `Date`/`DateTime` and the 2nd argument is a String. That's all the Canvas needs at the moment but I am happy to expand it to support all combinations

fedme: I have pushed a change to support the other args ordering you mentioned @smn https://github.com/turnhub/expression/pull/188/commits/4adbc14ea2c1917660339048234e967cc558c936

PR #194, opened 2024-05-08
fix lazy eval of ors

smn:
Finally tracked down why ors were greedy and returning the last value instead of the first.


smn:
@rudigiesler tagging you on this one too as you'd reported strange behaviour in `or`s - this was likely it.

fedme (approved):
Nice fix!

PR #201, opened 2024-06-17
guard against nils when comparing date values

smn:
Given the flexibility for Expressions, it's possible for `nil` values to be given to a date/datetime comparison function.
When that happens, return a reasonable error rather than crashing on `Date.compare` not accepting nils.

### DIFF
--- a/lib/expression/callbacks/standard.ex
+++ b/lib/expression/callbacks/standard.ex
@@ -1037,6 +1037,7 @@ defmodule Expression.Callbacks.Standard do
   @expression_doc expression: "has_date(\"the date is 15/01/2017 05:50\")",
                   result: %{
                     "__value__" => true,
+                    "match" => ~U[2017-01-15 05:50:00Z],
                     "date" => ~D[2017-01-15],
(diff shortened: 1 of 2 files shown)


benonymus (approved):

lib/expression/callbacks/standard.ex
```diff
@@ -1053,55 +1059,124 @@ defmodule Expression.Callbacks.Standard do
...
+                    "test" => ~D[2017-01-15]
+                  }
   def has_date_eq(ctx, expression, date_string) do
     [expression, date_string] = eval_args!([expression, date_string], ctx)
     found_date = DateHelpers.extract_dateish(expression)
     test_date = DateHelpers.extract_dateish(date_string)
-    # Future match result: found_date
-    found_date == test_date
+
+    with {:ok, match} <- date_compare(found_date, test_date) do
```
benonymus: I agree withe credo regarding not using `with` for a single check, `case` is sufficient.

smn: thanks!! fixing

PR #204, opened 2024-06-30
fix: kernel operations with numbers as strings

santiagocardo:
Fix kernel operations (`==, !=, +, -, *, /, >, >=, <, <=`) with numbers being passed as strings


smn (approved):
Thank you! Let's make sure we also document this behaviour properly.

PR #206, opened 2024-07-01
do not crash on extracting date-ish things

smn:
If an integer was supplied it would hit an unsupported code path rather than returning `nil` like it should.


santiagocardo (approved):

smn:
thanks @santiagocardo !

PR #208, opened 2024-07-03
[sc-34763] attempt to reproduce list index miss

smn:


santiagocardo:
Thanks for reproducing it while I was not online @smn 🙂
I have addressed the issue in #209

PR #209, opened 2024-07-03
fix(context): parsing zero as string

santiagocardo:
Fix context parsing with zero as a string.

### DIFF
--- a/lib/expression/context.ex
+++ b/lib/expression/context.ex
@@ -49,14 +49,15 @@ defmodule Expression.Context do
     {key, evaluate!(value, opts)}
   end
 
-  # Prevent implictly converting numbers starting with a zero
+  # Implictly convert the string "0" as a number
   defp iterate({key, "0"}, _opts), do: {key, 0}
 
+  # Prevent implictly converting numbers starting with a zero
   defp iterate({key, "0" <> value}, opts) when is_binary(value) do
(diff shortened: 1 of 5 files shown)


lib/expression/context.ex
```diff
@@ -50,11 +50,13 @@ defmodule Expression.Context do
...
 
   # Prevent implictly converting numbers starting with a zero
+  defp iterate({key, "0"}, _opts), do: {key, 0}
+
   defp iterate({key, "0" <> value}, opts) when is_binary(value) do
     if String.match?("0" <> value, ~r/^[0-9]/) do
       {key, "0" <> value}
     else
-      iterate({key, value}, opts)
+      iterate({key, "0" <> value}, opts)
```
fedme: Isn't this calling itself recursively using the same exact arguments? I am confused

santiagocardo: Yeah, nice catch. I'm fixing it now, thanks!
It went out as part of some debugging 🙃

santiagocardo: Done in [8d52db4](https://github.com/turnhub/expression/pull/209/commits/8d52db470525e523c66560fcf7c6c9eeb34dd431)

fedme (approved):
Nice catch thanks!

smn (approved):

PR #215, opened 2024-08-08
fix(eval): parse_number()

santiagocardo:
Fix `Expression.Eval.parse_number/1` when handling strings combined with numbers and punctuations.

### DIFF
--- a/test/expression/eval_test.exs
+++ b/test/expression/eval_test.exs
@@ -115,7 +115,7 @@ defmodule Expression.EvalTest do
 
   describe "lambdas" do
     test "with map" do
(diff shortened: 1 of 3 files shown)


fedme:
Thanks @santiagocardo! I have pushed a small change to try to parse the number as Integer first.

fedme (approved):
Thank you @santiagocardo!

test/expression/eval_test.exs
```diff
@@ -138,6 +138,16 @@ defmodule Expression.EvalTest do
                Eval.eval!(ast, %{"foo" => [1, 2, 3]})
     end
 
+    test "with kerner operators" do
```
fedme:
```suggestion
    test "with kernel operators" do
```

PR #219, opened 2024-08-28
`has_number()` : handle `nil` and case when expression contain just a number

Djack1010:
`has_number()` fails whether the input value is `nil`, such as:
```
iex> Expression.evaluate("@has_number(value)", %{"value" => nil})
** (FunctionClauseError) no function clause matching in String.replace/4
```

This PR extends `extract_numberish()` to handle `nil` input and cases when the expression contains just a number.
Examples that would fail without the edit:
```
iex> Expression.evaluate("The answer is @has_number(answer)", %{"answer" => nil})
{:ok, ["The answer is ", false]}
iex> Expression.evaluate("The answer is @has_number(answer)", %{"answer" => "42"})
{:ok, ["The answer is ", true]}
```

This is particularly needed because Expression does not implement lazy evaluation of the `and` boolean operator. As a result, we get errors every time a `nil` value is passed to `has_number()`, and we cannot fix it by simply adding a null check beforehand.
For instance, we get error here :
```
iex> Expression.evaluate("@and(isstring(value), has_number(value))", %{"value" => nil})
** (FunctionClauseError) no function clause matching in String.replace/4
```
The issue occurs because the entire expression is always evaluated, even if `isstring(nil)` results in `false` and the entire `and` would result in `false` anyway.

### DIFF
--- a/lib/expression/callbacks/standard.ex
+++ b/lib/expression/callbacks/standard.ex
@@ -1262,6 +1262,9 @@ defmodule Expression.Callbacks.Standard do
     !!group
   end
 
+  defp extract_numberish(nil), do: nil
+  defp extract_numberish(value) when is_number(value), do: value
+
   defp extract_numberish(expression) do
     with [match] <-
            Regex.run(~r/([0-9]+\.?[0-9]*)/u, replace_arabic_numerals(expression), capture: :first),
--- a/lib/expression/v2/callbacks/standard.ex
+++ b/lib/expression/v2/callbacks/standard.ex
@@ -1001,6 +1001,7 @@ defmodule Expression.V2.Callbacks.Standard do
     !!group
   end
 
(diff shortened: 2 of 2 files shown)


Djack1010:
**_I have read the CLA Document and I hereby sign the CLA_**

Djack1010:
**recheck**

Djack1010:
@smn This should fix the problem, but I'm not very familiar with this codebase, let me know if it makes sense. I can also add a few tests, but I'm not sure which file is best for adding tests for `has_number()`, as I couldn't find any existing ones.

lorenzosinisi:
Nice one @Djack1010, this goes along with another one https://github.com/turnhub/expression/pull/220 which could work in pair!

smn (approved):

smn (approved):

PR #221, opened 2024-09-11
Fix: handle nil expression value on `has_pattern` function

davidkhk:
`Expression.Callbacks.Standard.has_patter/3` fails when the `expression` arg value is `nil`, such as:

```
iex> Expression.evaluate_block!("has_pattern(nil, \"buy (\\w+)\")", %{}, Expression.Callbacks.Standard)
** (FunctionClauseError) no function clause matching in String.trim/1
```

This PR handles with `nil` expression values returning `false` for such cases.

### DIFF
--- a/lib/expression/callbacks/standard.ex
+++ b/lib/expression/callbacks/standard.ex
@@ -1492,7 +1492,7 @@ defmodule Expression.Callbacks.Standard do
   def has_pattern(ctx, expression, pattern) do
     [expression, pattern] = eval_args!([expression, pattern], ctx)
 
-    with false <- is_nil(expression),
+    with true <- is_binary(expression),
          {:ok, regex} <- Regex.compile(String.trim(pattern), "i"),
          [[_first | _remainder]] <- Regex.scan(regex, String.trim(expression), capture: :all) do
       # Future match result: first
--- a/lib/expression/callbacks/standard.ex
+++ b/lib/expression/callbacks/standard.ex
@@ -1489,14 +1489,11 @@ defmodule Expression.Callbacks.Standard do
   @expression_doc expression: "has_pattern(\"Buy cheese please\", \"buy (\\w+)\")", result: true
   @expression_doc expression: "has_pattern(\"Sell cheese please\", \"buy (\\w+)\")", result: false
   @expression_doc expression: "has_pattern(nil, \"buy (\\w+)\")", result: false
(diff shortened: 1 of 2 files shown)


davidkhk:

davidkhk:
recheck

nathanbegbie:
@davidkhk Thanks! Perhaps just add a test to `test/expression_test.exs` and we should be good to go? TBH I rarely touch this code, so need to defer to @smn or @santiagocardo on actually approving

nathanbegbie:
Not sure that I understand what we have to actually do to get the CLA thingo to pass 🤔 

Screenshot 2024-09-11 at 11 27 51

davidkhk:
> @davidkhk Thanks! Perhaps just add a test to `test/expression_test.exs` and we should be good to go? TBH I rarely touch this code, so need to defer to @smn or @santiagocardo on actually approving

Awesome, tests added. Will ask them to review, thanks!

davidkhk:
recheck

> Not sure that I understand what we have to actually do to get the CLA thingo to pass 🤔
> 
> Screenshot 2024-09-11 at 11 27 51

Me neither, commenting `recheck` doesn't seem to work.

nathanbegbie:
@smn @santiagocardo nudge nudge!

lib/expression/callbacks/standard.ex
```diff
@@ -1488,10 +1488,12 @@ defmodule Expression.Callbacks.Standard do
   """
   @expression_doc expression: "has_pattern(\"Buy cheese please\", \"buy (\\w+)\")", result: true
   @expression_doc expression: "has_pattern(\"Sell cheese please\", \"buy (\\w+)\")", result: false
+  @expression_doc expression: "has_pattern(nil, \"buy (\\w+)\")", result: false
   def has_pattern(ctx, expression, pattern) do
     [expression, pattern] = eval_args!([expression, pattern], ctx)
 
-    with {:ok, regex} <- Regex.compile(String.trim(pattern), "i"),
+    with false <- is_nil(expression),
```
santiagocardo: @davidkhk following the [Good and bad Elixir](https://keathley.io/blog/good-and-bad-elixir.html) reference, we can maybe change this a bit to "State what you want, not what you don’t"

```suggestion
    with true <- is_binary(expression),
```

santiagocardo (approved):
LGTM! Thanks for your contribution @davidkhk 🙂

PR #234, opened 2024-12-03
Fix: handle nil expression in `evaluate_as_string`

maryamaljanabi:
`evaluate_as_string/3` method fails when `expression` is `nil`. There are cases when `expression` is sent as `nil` since this method is called from `flow_runner` and various methods on `engage` backend where this case happens.

This PR handles the `nil` case in this method and adds the required test.

### DIFF
--- a/lib/expression.ex
+++ b/lib/expression.ex
@@ -121,14 +121,16 @@ defmodule Expression do
 
   def evaluate_as_string!(expression, context \\ %{}, mod \\ Expression.Callbacks) do
     case expression do
-      nil -> ""
+      nil ->
+        ""
+
       _ ->
         expression
         |> parse!
         |> Eval.eval!(Context.new(context), mod)
         |> Eval.default_value(handle_not_found: true)
         |> stringify()
-      end
+    end
   end
 
   def evaluate_as_boolean!(expression, context \\ %{}, mod \\ Expression.Callbacks) do
--- a/lib/expression.ex
+++ b/lib/expression.ex
@@ -119,18 +119,16 @@ defmodule Expression do
     |> Eval.default_value()
   end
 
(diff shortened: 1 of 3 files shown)


maryamaljanabi:

maryamaljanabi:
recheck

lib/expression.ex
```diff
@@ -120,11 +120,15 @@ defmodule Expression do
   end
 
   def evaluate_as_string!(expression, context \\ %{}, mod \\ Expression.Callbacks) do
```
smn:
```suggestion
  def evaluate_as_string!(expression, context \\ %{}, mod \\ Expression.Callbacks)
  def evaluate_as_string!(nil, _context, mod), do: ""
  def evaluate_as_string!(expression, context, mod) do
```

Thanks @maryamaljanabi the change is correct but could we use pattern matching here?

smn (approved):
Approving because it's correct :)

smn:
could you also attend to the formatting change required @maryamaljanabi  ?

smn:
@maryamaljanabi don't worry about the CLA assistant, I can merge it on your behalf. That thing is a bit buggy at times.

lib/expression.ex
maryamaljanabi: Thanks a lot, I'll make an update

maryamaljanabi:
@smn I've made the update and tests are passing. I have also tested that these changes will solve the `nil` bug with Santiago's suggestion on Slack. The CLA is still failing though, is there something specific I should commit for it to pass?

smn:
no I think this is great, the fact that the CLA is failing is strange. I made a change and now a second CLA thing is passing -- hopefully that'll solve it for the next PR. Please go ahead and merge!

PR #235, opened 2024-12-09
Fix/Add spec to `evaluate_as_string`

maryamaljanabi:
The current dialyzer in our main repo is failing after the latest release. Perhaps adding the spec will help fix the issue since the method has multiple clauses now.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -12,7 +12,7 @@ jobs:
       matrix:
         elixir: ["1.15"]
         otp: ["26"]
-        cache_version: ["1"]
+        cache_version: ["2"]
 
     steps:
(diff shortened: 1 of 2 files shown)


maryamaljanabi:
The `engage` repo's dialyzer is failing after updating to the latest versions for `expression` and `flow_runner` [here](https://github.com/turnhub/engage/actions/runs/12166421845/job/33933512753?pr=7797). The dialyzer and tests for both `expression` and `flow_runner` didn't show any errors. I'm guessing it could happen because the spec is missing so I added it in this PR. If you think the issue could be related to something else, please let me know. The failing pipeline only fails in github and not locally.

smn (approved):
Looks good, thanks @maryamaljanabi !

PR #238, opened 2024-12-11
Fix `evaluate_as_string` spec

maryamaljanabi:
Adding `Number.t()` to the expression types for `evaluate_as_string` spec. The reason for this is a failure in the main repo when the spec for `evaluate_as_string` was added. Also the `parser` allows string and number types ` @spec parse!(String.t() | Number.t()) :: Keyword.t() `

Adding a release `2.41.4`

### DIFF
--- a/lib/expression.ex
+++ b/lib/expression.ex
@@ -120,7 +120,7 @@ defmodule Expression do
   end
 
   @spec evaluate_as_string!(
-          String.t() | nil,
+          any(),
           map(),
           module()
         ) :: String.t()
--- a/lib/expression.ex
+++ b/lib/expression.ex
@@ -120,7 +120,7 @@ defmodule Expression do
   end
 
   @spec evaluate_as_string!(
-          any(),
+          String.t() | Number.t() | nil,
           map(),
           module()
         ) :: String.t()
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ by adding `expression` to your list of dependencies in `mix.exs`:
 ```elixir
 def deps do
   [
-    {:expression, "~> 2.41.3"}
+    {:expression, "~> 2.41.4"}
   ]
 end
 ```
--- a/mix.exs
+++ b/mix.exs
@@ -1,7 +1,7 @@
 defmodule Expression.MixProject do
   use Mix.Project
 
-  @version "2.41.3"
+  @version "2.41.4"
 
   def project do
     [


lib/expression.ex
```diff
@@ -120,7 +120,7 @@ defmodule Expression do
   end
 
   @spec evaluate_as_string!(
-          String.t() | nil,
+          any(),
```
smn: This may solve the typespec issue but whatever we're passing in must be string or a nil, otherwise the `parse!` on line 133 will fail.

smn (dismissed):
I think the problem here (wherever this is surfacing) is likely more wherever this function is being called and that allowing non-string, non-nil values. The `parse!` function that the expression is supplied to afaik only accepts a string. Fixing the typespec here likely fixes the symptom but not the cause. Where is this problem surfacing and can we start investigating there as well ?

lib/expression.ex
maryamaljanabi: The `parse` that's used in `evaluate_as_string` accepts a string or a number:
https://github.com/turnhub/expression/blob/6e35e4b9e3d0f59d46a1d7fc67195bdba5189667/lib/expression.ex#L86
The `parse_expression` is the one that only accepts a string:
https://github.com/turnhub/expression/blob/6e35e4b9e3d0f59d46a1d7fc67195bdba5189667/lib/expression.ex#L51

The update triggered a dialyzer error in the main repo somewhere that was passing a value for the expression and another method that's consuming that value and expecting it to be `binary() | Number.t()`. I can add a spec identical to the `parse` one with nil like `String.t() | Number.t() | nil`?

Since this is a public repo, not sure if I should paste links to where it's failing. I'll add it on Slack and tag you.

maryamaljanabi: I replaced `@any` with `String.t() | Number.t() | nil` to match the parser and also allow nil which we recently supported.

maryamaljanabi:
Hi @santiagocardo would love some feedback here. I noticed you introduced the ability for parsers to parse numbers in this line
https://github.com/turnhub/expression/blob/6e35e4b9e3d0f59d46a1d7fc67195bdba5189667/lib/expression.ex#L86

Does this mean the `evaluate_as_string` method should also be able to parse numbers? We are discussing that in a slack thread and just wanted to confirm that it's okay to change the type of the input to this method from `String.t() | nil` to `String.t() | Number.t() | nil`. I added `String.t() | nil` a few days ago thinking numbers shouldn't be there but noticed it's there in the parser and also dialyzer is complaining in our codebase in some areas I highlighted in the slack thread when I added that.

santiagocardo:
> Hi @santiagocardo would love some feedback here. I noticed you introduced the ability for parsers to parse numbers in this line
> 
> https://github.com/turnhub/expression/blob/6e35e4b9e3d0f59d46a1d7fc67195bdba5189667/lib/expression.ex#L86
> 
> Does this mean the `evaluate_as_string` method should also be able to parse numbers? We are discussing that in a slack thread and just wanted to confirm that it's okay to change the type of the input to this method from `String.t() | nil` to `String.t() | Number.t() | nil`. I added `String.t() | nil` a few days ago thinking numbers shouldn't be there but noticed it's there in the parser and also dialyzer is complaining in our codebase in some areas I highlighted in the slack thread when I added that.

Yes, @maryamaljanabi `Expression.evaluate_as_string!/1` should also accept numbers because `Expression.parse!/1` already accepts numbers:
https://github.com/turnhub/expression/blob/6e35e4b9e3d0f59d46a1d7fc67195bdba5189667/lib/expression.ex#L86

So yes, changing that function's type to `String.t() | Number.t() | nil` is okay. Thanks for fixing it!

lib/expression.ex
santiagocardo: @smn I have changed `Expression.parse!/1` to accept numbers in #129. So `Expression.evaluate_as_string!/1` should also accept numbers. In that case, the spec should be `String.t() | Number.t() | nil` instead of `any()` as @maryamaljanabi did in [0f05eaa](https://github.com/turnhub/expression/pull/238/commits/0f05eaa560143cd7ca7370bf090eee3806fd35f8).

santiagocardo (approved):

maryamaljanabi:
@smn I can't merge due to your open review request and it feels disrespectful to dismiss it. Santiago approved and you can see the discussion above. If you see this, please remove the merge block so I can merge this and go on with resolving the bug fix in the main repo as this is needed there. Thanks!

smn:
@maryamaljanabi thank you for addressing this and thanks @santiagocardo for weighing in on the number change!

smn (approved):

PR #247, opened 2025-02-26
Move Turn callbacks to OSS release

smn:
There's no good reason to keep these out of this repo.
The remaining ones in Turn either touch the turn code base directly or rely on things like Timezone analysis for datetime calculations or use the Peach library for some text wrangling. Neither of which I'm keen to pull into this library as a dependency.


smn:
@rudigiesler FYI

PR #250, opened 2025-03-14
Fix: handle nil values on date.compare/2 on eval

davidkhk:
`Date.compare/2` fails when one of the date is `nil`.

### DIFF
--- a/lib/expression/eval.ex
+++ b/lib/expression/eval.ex
@@ -203,6 +203,11 @@ defmodule Expression.Eval do
   # Support comparing a `Date`/`DateTime` value with an ISO8601 date/datetime string
   # credo:disable-for-next-line Credo.Check.Refactor.CyclomaticComplexity
   def op(operator, a, b)
+      when operator in [:=, :==, :!=, :<, :<=, :>, :>=] and (is_nil(a) or is_nil(b)) do
+    false
+  end
+
+  def op(operator, a, b)
       when operator in [:=, :==, :!=, :<, :<=, :>, :>=] and
              (((is_struct(a, Date) or is_struct(a, DateTime)) and is_binary(b)) or
                 (is_binary(a) and (is_struct(b, Date) or is_struct(b, DateTime)))) do
@@ -215,15 +220,10 @@ defmodule Expression.Eval do
       end
 
     comparison_result =
-      cond do
-        is_nil(date_a) or is_nil(date_b) ->
-          nil
-
-        is_struct(date_a, Date) ->
-          Date.compare(date_a, date_b)
-
-        true ->
-          DateTime.compare(date_a, date_b)
+      if is_struct(date_a, Date) do
+        Date.compare(date_a, date_b)
+      else
+        DateTime.compare(date_a, date_b)
       end
 
     case {operator, comparison_result} do
--- a/lib/expression/eval.ex
+++ b/lib/expression/eval.ex
@@ -215,10 +215,15 @@ defmodule Expression.Eval do
       end
 
     comparison_result =
-      if is_struct(date_a, Date) do
-        Date.compare(date_a, date_b)
-      else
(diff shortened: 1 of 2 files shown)


davidkhk:
@fedme for some context, this is the [issue](https://linear.app/turnio/issue/PROD-1485/messages-in-dd-mm-aaaa-format-are-evaluated-as-date) and [sentry error](https://praekeltorg-pbc.sentry.io/issues/6268536463/?project=4504797031038976).

lib/expression/eval.ex
```diff
@@ -215,10 +215,15 @@ defmodule Expression.Eval do
       end
 
     comparison_result =
-      if is_struct(date_a, Date) do
-        Date.compare(date_a, date_b)
-      else
-        DateTime.compare(date_a, date_b)
+      cond do
+        is_struct(date_a, Date) and is_nil(date_b) ->
```
fedme:
```suggestion
        is_nil(date_a) or is_nil(date_b) ->
```

I would suggest changing the check to this so that we can match nils in both positions

fedme:
Thanks @davidkhk! Could you add some tests too if possible?

davidkhk:
> Thanks @davidkhk! Could you add some tests too if possible?

Thanks for the review! The argument itself was `nil` so I created a new clause for `op/3` handling the case. Also added tests.

test/expression/eval_test.exs
```diff
@@ -399,4 +399,52 @@ defmodule Expression.EvalTest do
...
+      # Direct tests of the op function for nil handling
+      date = ~D[2024-01-01]
+      datetime = ~U[2024-01-01 10:40:50.277482Z]
+      date_str = "2024-01-01"
+      datetime_str = "2024-01-01T10:40:50.277482Z"
+
+      # Date with nil string tests
+      assert false == Expression.Eval.op(:>, date, nil)
+      assert false == Expression.Eval.op(:<, date, nil)
+      assert false == Expression.Eval.op(:==, date, nil)
```
santiagocardo, line 414: @davidkhk Can we please also add a case comparing `nil` with `nil`?
```elixir
      assert false == Expression.Eval.op(:==, nil, nil)
      assert false == Expression.Eval.op(:=, nil, nil)
```

lib/expression/eval.ex
```diff
@@ -202,6 +202,11 @@ defmodule Expression.Eval do
 
   # Support comparing a `Date`/`DateTime` value with an ISO8601 date/datetime string
   # credo:disable-for-next-line Credo.Check.Refactor.CyclomaticComplexity
+  def op(operator, a, b)
+      when operator in [:=, :==, :!=, :<, :<=, :>, :>=] and (is_nil(a) or is_nil(b)) do
```
joaothallis, line 206: Should we return `false` when both are `nil`? Is it a possible case?

test/expression/eval_test.exs
joaothallis: Should we turn false when both are `nil`?

joaothallis (approved):

test/expression/eval_test.exs
```diff
@@ -399,4 +399,59 @@ defmodule Expression.EvalTest do
...
+
+      # DateTime string with nil
+      assert false == Expression.Eval.op(:>, datetime_str, nil)
+      assert false == Expression.Eval.op(:<, datetime_str, nil)
+      assert false == Expression.Eval.op(:==, datetime_str, nil)
+
+      # Comparing nil with nil values
+      assert false == Expression.Eval.op(:>, nil, nil)
+      assert false == Expression.Eval.op(:<, nil, nil)
+      assert false == Expression.Eval.op(:==, nil, nil)
```
santiagocardo, line 452: @davidkhk I think this should be `true` as both are nil values and so they are equal.

PR #251, opened 2025-03-17
Fix op/3 clause and test

davidkhk:
Fix previously merged [PR](https://github.com/turnhub/expression/pull/250), fixing op/2 clause and test.

### DIFF
--- a/lib/expression/eval.ex
+++ b/lib/expression/eval.ex
@@ -201,7 +201,6 @@ defmodule Expression.Eval do
     do: Date.compare(a, b) == :eq
 
   # Support comparing a `Date`/`DateTime` value with an ISO8601 date/datetime string
-  # credo:disable-for-next-line Credo.Check.Refactor.CyclomaticComplexity
   def op(operator, a, b) when operator in [:=, :==] and is_nil(a) and is_nil(b) do
     true
   end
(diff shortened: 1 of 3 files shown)


lib/expression/eval.ex
```diff
@@ -202,6 +202,10 @@ defmodule Expression.Eval do
 
   # Support comparing a `Date`/`DateTime` value with an ISO8601 date/datetime string
   # credo:disable-for-next-line Credo.Check.Refactor.CyclomaticComplexity
+  def op(operator, a, b) when operator == :== and is_nil(a) and is_nil(b) do
```
santiagocardo: Let's include `:=` for backward compatibility.
```suggestion
  def op(operator, a, b) when operator in [:=, :==] and is_nil(a) and is_nil(b) do
```

test/expression/eval_test.exs
```diff
@@ -447,9 +447,9 @@ defmodule Expression.EvalTest do
       assert false == Expression.Eval.op(:==, datetime_str, nil)
 
       # Comparing nil with nil values
+      assert true == Expression.Eval.op(:==, nil, nil)
```
santiagocardo, line 451:
```suggestion
      assert true == Expression.Eval.op(:=, nil, nil)
      assert true == Expression.Eval.op(:==, nil, nil)
```

santiagocardo (approved):
LGTM!

PR #256, opened 2025-03-30
feat: add date validation to `date` expression

joaothallis:
This PR modifies the `date` function in both `Expression.Callbacks.Standard` and `Expression.V2.Callbacks.Standard` modules to incorporate input validation and proper error handling when constructing a `Date` struct. Previously, the functions directly constructed a date struct without verifying the inputs, which could lead to improper or invalid date values.

Changes include:
- Using a `with` construct to improve readability and streamline the validation process.
- Adding checks to ensure `year`, `month`, and `day` are integers using `is_integer/1`.
- Creating a date using `Date.new/3`, which ensures the date is valid and returns a tuple with `{:ok, date}` or an error.
- Incorporating an `else` clause to handle invalid input scenarios by invoking `Expression.error/1` with the message "Invalid date".

This feature enhances robustness by preventing the construction of invalid date structs and providing better feedback in case of errors.

related to PROD-1497

### DIFF
--- a/lib/expression/callbacks/standard.ex
+++ b/lib/expression/callbacks/standard.ex
@@ -90,16 +90,14 @@ defmodule Expression.Callbacks.Standard do
   def date(ctx, year, month, day) do
     [year, month, day] = eval_args!([year, month, day], ctx)
 
-    fields = [
-      calendar: Calendar.ISO,
-      year: year,
-      month: month,
-      day: day,
-      time_zone: "Etc/UTC",
-      zone_abbr: "UTC"
-    ]
-
-    struct(Date, fields)
+    with true <- is_integer(year),
+         true <- is_integer(month),
+         true <- is_integer(day),
+         {:ok, date} <- Date.new(year, month, day) do
+      date
+    else
+      _ -> Expression.error("Invalid date")
+    end
   end
 
   @doc """
--- a/lib/expression/callbacks/standard.ex
+++ b/lib/expression/callbacks/standard.ex
@@ -93,7 +93,7 @@ defmodule Expression.Callbacks.Standard do
                   result: %{
                     "__type__" => "expression/v1error",
                     "error" => true,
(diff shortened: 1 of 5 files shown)


smn (approved):
Happy for you to drop the v2 change, we're going to be removing that implementation entirely.

lib/expression/callbacks/standard.ex
```diff
@@ -87,19 +87,25 @@ defmodule Expression.Callbacks.Standard do
...
-    ]
-
-    struct(Date, fields)
+    with true <- is_integer(year),
+         true <- is_integer(month),
+         true <- is_integer(day),
+         {:ok, date} <- Date.new(year, month, day) do
+      date
+    else
+      _ -> Expression.error("Invalid date")
```
smn: Can we provide the supplied value as part of the error here? That'll put an author in a better position to diagnose the issue at hand.

smn:
```suggestion
      _ -> Expression.error("Invalid date: date(#{inspect(year)}, #{inspect(month)}, #{inspect(day)})") 
```
or something similar?

smn (approved):
Happy with the direction! Merge when you see fit :)

lib/expression/callbacks/standard.ex
joaothallis: Yes, great suggestion @smn, done in https://github.com/turnhub/expression/pull/256/commits/12d8dcb0059c292c89013df6f1a1f80615105b69

smn: cool, we'll just need to update the test to match I think

joaothallis: I added here: https://github.com/turnhub/expression/pull/256/commits/9f7fe414843beac7d567f3f9dd3dfdba3151d3a7

smn (approved):

PR #258, opened 2025-03-31
feat: return null value when invalid date

joaothallis:
This PR adds a `__value__` key to the error. It's to be able to separate what's the error message and what's the value returned by the function.

- dialyzer was not failing on CI but locally, this PR adds a new Elixir version that gets new dialyzer issues

related to PROD-1497

### DIFF
--- a/lib/expression.ex
+++ b/lib/expression.ex
@@ -170,7 +170,12 @@ defmodule Expression do
   """
   @spec error(message :: term) :: %{required(String.t()) => term}
   def error(message),
-    do: %{"__type__" => "expression/v1error", "error" => true, "message" => to_string(message)}
+    do: %{
+      "__type__" => "expression/v1error",
+      "error" => true,
+      "message" => to_string(message),
+      "__value__" => nil
+    }
 
   defdelegate prewalk(ast, fun), to: Macro
   defdelegate traverse(ast, acc, pre, post), to: Macro
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -1,8 +1,11 @@
 name: Elixir CI
 
 on:
-  push:
   pull_request:
+    types: [opened, synchronize]
+  push:
+    branches:
+      - develop
 
 jobs:
   build:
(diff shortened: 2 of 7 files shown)


lib/expression.ex
```diff
@@ -170,7 +170,12 @@ defmodule Expression do
   """
   @spec error(message :: term) :: %{required(String.t()) => term}
   def error(message),
-    do: %{"__type__" => "expression/v1error", "error" => true, "message" => to_string(message)}
+    do: %{
+      "__type__" => "expression/v1error",
```
santiagocardo, line 174: I'm not sure if this was a typo, or if it's the correct value here instead of `"__type__" => "expression/v1"`. @smn Do you know?

santiagocardo (commented):
@joaothallis Can we add some tests for this new attribute?

joaothallis:
> @joaothallis Can we add some tests for this new attribute?

There are these doctests: https://github.com/turnhub/expression/pull/258/files#diff-4c7c687fcc4f6018023ed8589c4a9c927b06743f9f53ae3cdb60d0ab507621f5 and this unit test:
https://github.com/turnhub/expression/pull/258/commits/fdcbbefcd39ed8f3cc88906577327af089beb0bf is it missing some case?

lib/expression.ex
joaothallis: It's that value in our documentation: 

1. https://whatsapp.turn.io/docs/build/expressions_reference#datetime_adddatetime-offset-unit
2. https://whatsapp.turn.io/docs/build/expressions_reference#is_errorvalue

santiagocardo:
> > @joaothallis Can we add some tests for this new attribute?
> 
> There are these doctests: https://github.com/turnhub/expression/pull/258/files#diff-4c7c687fcc4f6018023ed8589c4a9c927b06743f9f53ae3cdb60d0ab507621f5 and this unit test: [fdcbbef](https://github.com/turnhub/expression/commit/fdcbbefcd39ed8f3cc88906577327af089beb0bf) is it missing some case?

You are right @joaothallis 👍 Doc tests should cover this already.
About `"__type__" => "expression/v1error"`, we can fix it in a follow-up PR if it happens to be a typo.

santiagocardo (approved):

smn (approved):

PR #261, opened 2025-04-29
Add support for complex values in if statements

santiagocardo:
Add support for complex values with a `__value__` attribute in `if/3` expression callback.

Right now, it's failing because it does not support complex values returned by other expressions:
image


joaothallis (approved):
LGTM!

PR #263, opened 2025-05-06
[PROD-2052] Convert back Time type values to string on `parse!`

davidkhk:
Time inputs are automatically converted to Time structs. This PR converts such values back to strings on the function `parse!`.

### DIFF
--- a/lib/expression.ex
+++ b/lib/expression.ex
@@ -98,7 +98,7 @@ defmodule Expression do
     end
   end
 
-  @spec time_struct?(String.t() | Number.t() | Time.t()) :: boolean
+  @spec time_struct?(String.t() | Time.t()) :: boolean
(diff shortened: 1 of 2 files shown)


lib/expression.ex
```diff
@@ -96,6 +98,9 @@ defmodule Expression do
     end
   end
 
+  @spec time_struct?(String.t() | Number.t() | Time.t()) :: boolean
```
joaothallis: It's not possible to have a `Number` struct because it's converted to string before calling `parse/1`: https://github.com/turnhub/expression/blob/8e0ff5d6bcb69ca01bbefb8984c7e0f5e26ed35d/lib/expression.ex#L87-L87
```suggestion
  @spec time_struct?(String.t() | Time.t()) :: boolean
```

joaothallis (approved):
LGTM!

PR #274, opened 2025-08-20
Restrict boolean cast to lower case strings only

Djack1010:
Boolean-like strings are currently evaluated and cast to boolean values, but I think our evaluation is too permissive. At the moment, we allow any upper/lowercase variation of boolean strings (e.g. "False", "TRUE", "FaLsE", "TruE") to be cast to Elixir’s true and false.

I propose restricting this behavior so that only the lowercase strings "true" and "false" are cast to booleans. This way, the behavior perfectly matches Elixir’s reserved atoms true and false.

Note: We discovered this issue because the string "False" was consistently being converted to lowercase "false", due to the evaluation logic that normalizes all upper/lowercase combinations.


Djack1010:
I see there’s an option skip_context_evaluation? to handle these boolean-string values. I think it’s better to start using that option rather than restricting boolean casting for now, so I’m closing this PR.

for example, see this test from Expression.ContextTest
```
describe "context is parsed correctly when using the `skip_context_evaluation?` option" do
    test "string values in context that resemble booleans should not be parsed as booleans" do
      # By default (without the flag) boolean-ish string values as parsed as booleans
      assert %{"block" => %{"response" => true}} ==
               Context.new(%{
                 "block" => %{"response" => "True"}
               })

      # With the flag set to true they are kept as strings
      assert %{"block" => %{"response" => "True"}} ==
               Context.new(%{"block" => %{"response" => "True"}},
                 skip_context_evaluation?: true
               )

      assert %{"block" => %{"response" => "true"}} ==
               Context.new(
                 %{"block" => %{"response" => "true"}},
                 skip_context_evaluation?: true
               )
    end
```

PR #276, opened 2025-09-12
Tag functions with categories

smn:
This allows for logical grouping of functions in documentation and code helpers

Updates the `expression_docs()` macro to return the category in the tuple at position 2:

```elixir
{"count", :direct, "enum", ["term"],
   "Return the number of entries in a list, string, or a map.\n",
   [
     %{
       context: %{"nil_value" => nil},
       result: 0,
       expression: "count(nil_value)"
     },
     %{
       context: %{"map" => %{"foo" => "bar"}},
       result: 1,
       expression: "count(map)"
     },
     %{result: 3, expression: "count(\"zoë\")"},
     %{result: 3, expression: "count([1, 2, 3])"}
   ]}
```

### DIFF
--- a/lib/expression/callbacks/standard.ex
+++ b/lib/expression/callbacks/standard.ex
@@ -33,6 +33,7 @@ defmodule Expression.Callbacks.Standard do
   @doc """
   Return the number of entries in a list, string, or a map.
   """
+  @expression_category "enum"
   @expression_doc expression: "count([1, 2, 3])", result: 3
   @expression_doc expression: "count(\"zoë\")", result: 3
   @expression_doc expression: "count(map)", context: %{"map" => %{"foo" => "bar"}}, result: 1
@@ -53,6 +54,7 @@ defmodule Expression.Callbacks.Standard do
   to process them in smaller chunks.
 
   """
+  @expression_category "enum"
   @expression_doc doc: """
                   Split a large set of sentences into a smaller set of sentences.
                   """,
@@ -79,6 +81,7 @@ defmodule Expression.Callbacks.Standard do
   @doc """
   Defines a new date value
   """
+  @expression_category "date"
   @expression_doc doc: "Construct a date from year, month, and day integers",
                   expression: "date(year, month, day)",
                   context: %{
@@ -128,6 +131,7 @@ defmodule Expression.Callbacks.Standard do
   Specifying a negative offset results in date calculations back in time.
 
   """
+  @expression_category "date"
   @expression_doc doc: "Calculates a new datetime based on the offset and unit provided.",
                   expression: "datetime_add(datetime, offset, unit)",
                   context: %{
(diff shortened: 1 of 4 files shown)


Aratramba (approved):
I love it! Thanks for doing this! I explored this yesterday but couldn't make it work, so i went with the lists instead. But having this closer to the actual code is so much better! 🙌

lib/expression/callbacks/standard.ex
```diff
@@ -806,16 +849,19 @@ defmodule Expression.Callbacks.Standard do
   "You have 4.21 in your account"
   ```
   """
+  @expression_category "number"
   @expression_doc expression: "fixed(4.209922, 2, false)", result: "4.21"
   @expression_doc expression: "fixed(4000.424242, 4, true)", result: "4000.4242"
   @expression_doc expression: "fixed(3.7979, 2, false)", result: "3.80"
   @expression_doc expression: "fixed(3.7979, 2)", result: "3.80"
   @expression_doc expression: "fixed(0.0909, 2)", result: "0.09"
+  @expression_category "string"
```
Aratramba, line 858: This should be "string" for `fixed`

lib/expression/callbacks/standard.ex
```diff
@@ -1899,8 +1998,10 @@ defmodule Expression.Callbacks.Standard do
     end)
   end
 
+  @expression_category "number"
```
Aratramba, line 2001: should this be "string" instead of number?

lib/expression/callbacks/standard.ex
```diff
@@ -1915,6 +2016,7 @@ defmodule Expression.Callbacks.Standard do
   Will return whatever was supplied as is when the given
   argument is not a String.
   """
+  @expression_category "enum"
```
Aratramba, line 2019: should this be "string" for `parse_json`?

lib/expression/callbacks/standard.ex
```diff
@@ -317,6 +329,7 @@ defmodule Expression.Callbacks.Standard do
   Parse random dates and times with `strftime` patterns and return a DateTime value
   when it matches.
   """
+  @expression_category "date"
```
Aratramba, line 332: should this be "string" for `parse_datevalue`?

lib/expression/callbacks/standard.ex
smn: yeah seeing now it's been added twice, once for number once for string. Fixing in a follow-up release.

lib/expression/callbacks/standard.ex
smn: hm parse_float returns a float, so I'd say it needs to be number?

lib/expression/callbacks/standard.ex
smn: let's agree on how we categorise, do we categorise on the input or the output? I'm inclined to categorise on the output

Aratramba: Alright! As long as it's consistent and intentional  it's fine with me

PR #284, opened 2025-10-13
Eval default_value

shaunsephton:
Ensures that all enum values are converted to their default values (the value of the `__value__` key) when expressions are evaluated, thereby fully supporting enum values in all callbacks.


lib/expression/callbacks/eval_helpers.ex
```diff
@@ -3,13 +3,25 @@ defmodule Expression.Callbacks.EvalHelpers do
...
+  the value after evaluation is converted to its default value, which
+  is the value as is for non enum fields, and the `__value__` key for
+  enum fields.
   """
   @spec eval!(term, map) :: term
-  def eval!(ast, ctx) do
-    ast
-    |> Expression.Eval.eval!(ctx)
-    |> Expression.Eval.not_founds_as_nil()
+  def eval!(ast, ctx, with_defaults \\ true) do
```
smn, line 12: are we sure this doesn't have any unintended side effects?

shaunsephton: @smn @santiagocardo great question. I am not sure at all, this is an extremely risky PR. I am trusting the existing test harness to be exactly that, a harness to catch regressions. The problem however is that there are precious little tests and the tests that there are tend to focus on the happy path, nevermind checking the behavior of enums. I started adding enum tests for each callback yesterday but I soon realized it would take me days to complete that work, and even then I'd only really be checking enum themselves, anything else that rely on the eval not occurring that is not already tested in that manner would still regress. In the absence of proper coverage here, what would you propose I do to determine possible side effects?

shaunsephton: @smn @santiagocardo I'm busy adding some tests to the expressions lib to check behaviors with enum fields, but it seems like its a bit of a waste of time since I don't know what the expected/correct behaviour should be, and each of them is different probably with their own expected behaviours. here is an example test I'm adding:
```
@expression_doc expression: "date(year, month, day)",
                  context: %{
                    "year" => %{
                      "display" => "value for display key",
                      "value" => "value for value key",
                      "__value__" => 2025
                    },
                    "month" => %{
                      "display" => "value for display key",
                      "value" => "value for value key",
                      "__value__" => 1
                    },
                    "day" => %{
                      "display" => "value for display key",
                      "value" => "value for value key",
                      "__value__" => 15
                    }
                  },
                  # WHAT IS THE EXPECTED BEHAVIOUR HERE?
                  # - SHOULD IT RETURN AN ERROR, OR
                  # - USE THE VALUES IN THE __value__ KEY
                  # - OR USE THE VALUE AS IS
                  result: "AMBIGUOUS"
```

Before I continue down this path of highlighting the ambiguities, what is the expected behaviour for the majority of the expressions? If they are provided with enums for their arguments, should the use the value in the __value__ key, or should they use the enum as is, or should they return errors?

santiagocardo: @shaunsephton, first of all, I believe there is a bit of confusion here. Just to clarify, in the world of **contact fields**, an enum is basically a list of objects with the following schema: `[%{"display" => "option A", "value" => "OPTION_A"}]`. But in the world of **expressions**, the following schema is not necessarily an enum: `%{"display" => "option A", "value" => "OPTION_A", __value__ => "OPTION_A"}`, is more what we call a _complex_ object, whose value should always default to the attribute `__value__`. I know it's a bit confusing because in the world of Elixir, both structures use the _Enum_ protocol 😅💁🏻‍♂️

santiagocardo: Having said that, in the example you pointed out above, it should use the value in the `__value__` key. Are you getting errors when writing tests for expression callbacks that use these _complex_ objects?

santiagocardo: Asking because when writing a few test cases in [test/expression_test.exs](https://github.com/turnhub/expression/blob/f034381252db405ee01d803598cf65f2c9062f14/test/expression_test.exs), they all succeed:

```elixir
test "list with complex objects" do
  assert "medical_doctor" ==
            Expression.evaluate_as_string!("@occupation[0]", %{
              "occupation" => [
                %{
                  "__value__" => "medical_doctor",
                  "display" => "Medical Doctor",
                  "value" => "medical_doctor"
                }
              ]
            })

  assert "Medical Doctor" ==
            Expression.evaluate_as_string!("@occupation[0].display", %{
              "occupation" => [
                %{
                  "__value__" => "medical_doctor",
                  "display" => "Medical Doctor",
                  "value" => "medical_doctor"
                }
              ]
            })

  assert "medical_doctor" ==
            Expression.evaluate_as_string!(
              "@filter(occupation, &(&1 == \"medical_doctor\"))",
              %{
                "occupation" => [
                  %{
                    "__value__" => "lawyer",
                    "display" => "Lawyer",
                    "value" => "lawyer"
                  },
                  %{
                    "__value__" => "medical_doctor",
                    "display" => "Medical Doctor",
                    "value" => "medical_doctor"
                  }
                ]
              }
            )
end
```

shaunsephton: @santiagocardo the question is what is the expected behavior for the majority of the expressions? If they are provided with complex/enum objects for their arguments, should they use the value in the `__value__` key, or should they use the complex object as is, or should they return errors? None of the expressions currently use the `__value__`, and the majority of them currently will fail if provided with a complex/enum object. E.g. a `has_phrase` expression will fail when provided with a complex/enum. 

My previously included example test illustrates the question, i.e. what should happen in these scenarios where complex objects are provided to any expression:
```
@expression_doc expression: "date(year, month, day)",
                  context: %{
                    "year" => %{
                      "display" => "value for display key",
                      "value" => "value for value key",
                      "__value__" => 2025
                    },
                    "month" => %{
                      "display" => "value for display key",
                      "value" => "value for value key",
                      "__value__" => 1
                    },
                    "day" => %{
                      "display" => "value for display key",
                      "value" => "value for value key",
                      "__value__" => 15
                    }
                  },
                  # WHAT IS THE EXPECTED BEHAVIOUR HERE?
                  # - SHOULD IT RETURN AN ERROR, OR
                  # - USE THE VALUES IN THE __value__ KEY
                  # - OR SOMEHOW USE THE COMPLEX VALUE AS IS
                  result: "<WHAT SHOULD THE RESULT BE HERE???>"
```

santiagocardo: > If they are provided with complex/enum objects for their arguments, should they use the value in the `__value__` key?

@shaunsephton, so yeah, ideally, that should be the expected behavior. If that's not the case, then I think we should make it possible, as complex objects are meant to default to `__value__`. Right @smn?

lib/expression/callbacks/eval_helpers.ex
```diff
@@ -3,13 +3,25 @@ defmodule Expression.Callbacks.EvalHelpers do
...
+    result =
+      ast
+      |> Expression.Eval.eval!(ctx)
```
santiagocardo: Just a nitpick:
```suggestion
    result = Expression.Eval.eval!(ast, ctx)
```

lib/expression/callbacks/standard.ex
```diff
@@ -2505,7 +4026,14 @@ defmodule Expression.Callbacks.Standard do
...
   def is_error(ctx, value) do
-    case eval!(value, ctx) do
+    case eval!(
+           value,
+           ctx,
+           # Do not have enums describing errors be casted to the default
+           # value (which is nil), which would make it impossible to
+           # detect errors.
+           false
+         ) do
```
santiagocardo, line 4036: Just a nitpick
```suggestion
    # Do not have enums describing errors be casted to the default
    # value (which is nil), which would make it impossible to detect errors.
    with_defaults? = false

    case eval!(value, ctx, with_defaults?) do
```

lib/expression/callbacks/standard.ex
santiagocardo, line 1: @shaunsephton, thanks for adding all these examples using _complex_ args! I just have a question regarding public docs: should we only add _complex_ args examples to certain expressions to avoid flooding the [public docs](https://whatsapp.turn.io/docs/build/expressions_reference) with too much info/examples?

santiagocardo (commented):
Thanks, @shaunsephton 🙌 It looks like it's working — most tests are passing!

I tried locally, but it looks like there is a single failing test in [autodoc_test.exs](https://github.com/turnhub/expression/blob/77670b865cdc4178af3f8ea75b20c46e25a76a3a/test/expression/autodoc_test.exs). Also, there is a Dialyzer issue.

lib/expression/callbacks/standard.ex
shaunsephton: This doctest approach is severely lacking IMHO, but since this is the way the original implementation was designed I'm sticking with the convention. For me more docs/examples are always better, but regardless, I have already wasted a huge amount of time adding these tests that should have been present in place in the first place and don't want to now waste even more time creating a separate test suite. If there is no concrete technical reason why its not fine as is please approve this PR so we can proceed on other urgent tasks. Otherwise I will delegate reworking of these suites to the original author.

shaunsephton:
> Thanks, @shaunsephton 🙌 It looks like it's working — most tests are passing!
> 
> I tried locally, but it looks like there is a single failing test in [autodoc_test.exs](https://github.com/turnhub/expression/blob/77670b865cdc4178af3f8ea75b20c46e25a76a3a/test/expression/autodoc_test.exs). Also, there is a Dialyzer issue.

Thanks, I've fixed the autodoc tests by simply duplicating it. Nothing in that module is documented in any meaningful way, so I have no clue why those specific tests are duplicated like that. Since its not documented I'm assuming this is fine.

santiagocardo (approved):

lib/expression/callbacks/standard.ex
santiagocardo: Cool, I also agree that these are concerns for a separate PR. Let's merge this one as is now 👍

PR #285, opened 2025-10-14
Add test illustrating ambiguity of evaluate_block vs evaluate_as_string

shaunsephton:


test/expression_test.exs
```diff
@@ -614,4 +614,19 @@ defmodule ExpressionTest do
...
   end
+
+  describe "evaluate_block! vs evaluate_as_string!" do
+    test "todo" do
+      # It is unclear to me what the right behaviour is here when an error occurs, specifically:
+      # - Should the result for evaluate_block! be an error map, or should it be the empty string?
+      # - Should the result for evaluate_as_string! be an error map, or should it be the empty string?
+      # - In either case, should the result for evaluate_block! and evaluate_as_string! be the same?
+      #   If not why not.
+      # - If it should be the empty string, how is that useful? Should it not then rather be the message
```
smn: It's been a while that I looked at this but if I recall correctly, the reason we have evaluate_as_string is to ensure that whatever we get out should be presentable to an end user via some kind of text message. 

evaluate block should return the raw data structure, I would expect that to be an error map. Note that we're not prefixing it with an `@` here for this one.

The `@` indicates that this is a template interpolation of an evaluated expression. And so a test that's more representative of the usage here would be something like:

```
Expression.evaluate_as_string!("Hi there, these are your options: @chunk_every(nil, 2)")
```

The `Expression.evaluate_block!/2` is used in places in the flow runner where an expression needs to be evaluated to an actual typed value instead of a templated string interpolated value, so stuff like:

```
Expression.evaluate_block!("contact.age > 18")
```

should return a boolean, not a string `"true"`

shaunsephton: Thanks @smn , I've updated the tests in the PR as per your feedback, specifically it now checks that:
- `evaluate_block!` returns a raw data structure.
- `evaluate_as_string!` returns a string.

This is fine, however there is still my open question re. whether or not `evaluate_as_string!` should return the empty string if an error occured. How is that useful? Should it not then rather be the message as sourced from the error map?

smn: yes, it should return the error as sourced from the map. The `evaluate_as_string!` will default to the `__value__` key of any value returned if it exists and then convert that to a string. The error we're returning as `__value__` as `nil` by default and the `.message` has the error message. The idea here is to allow folks using expressions to introspect whether a value returned is an error or not and then deciding how to handle that. Experience informs us that developers will _rarely_ put in an error that's actually useful for sending through to the end-user via a text messaging channel and so as a safety precaution the `evalute_as_string!` defaults to nil.

Happy to debate whether this is a good design decision but it's informed by situations where people started accidentally receiving things like `"Your registration is ERROR: TOKEN XXX IS INVALID"` which was worse than sending a nil value.

smn (dismissed):

test/expression_test.exs
shaunsephton: Thank you, that clarifies it @smn . I have now added another test explicitly describing these differences and the need for the empty string, if you can review again and approve if happy please.

lib/expression/autodoc.ex
```diff
@@ -167,6 +169,40 @@ defmodule Expression.Autodoc do
...
+  way to how the autodoc generated tests work. This is useful when creating unit
+  tests (instead of doc tests) or when trying to debug doc tests.
+
+  Note that there is a subtle difference here however in that for each eval
+  (evaluate_block! and evaluate_as_string!) an explicit expected result should be
+  provided (instead of result being stringified and used to assert the result of
+  evaluate_as_string!). This is to more clearly separate the two tests and ensure that
+  the expected string result is explicitly defined as exactly what is expected.
+  """
+  def test_expression(
```
smn: can we perhaps house this in `test/expression_test.exs` as that's where it's being used and already imports the required asserts

smn (approved):
thank you!

PR #296, opened 2025-11-13
Add boolean type support to parse! function

davidkhk:
This PR fixes a `FunctionClauseError` that occurs when boolean values (`true` or `false`) are passed to `Expression.evaluate_as_string!/3`.

Currently, the `parse!/1` function handles several primitive types by converting them to strings before parsing:
- Numbers (integers and floats)
- Time structs
- Nil values

However, it lacks a clause for boolean values, causing crashes when booleans are passed directly to evaluation functions like `evaluate_as_string!`. This is inconsistent behavior, as booleans are a common primitive type that should be stringifiable just like numbers.

### DIFF
--- a/lib/expression.ex
+++ b/lib/expression.ex
@@ -83,8 +83,9 @@ defmodule Expression do
     String.replace(expression, ~r/@([a-z]+)(\(|\.)?/i, "@@\\g{1}\\g{2}")
   end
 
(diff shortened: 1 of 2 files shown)


joaothallis (approved):
Looks good, thanks @davidkhk!

PR #304, opened 2026-01-21
[prod-4431][bugfix] add typecasting to url_encode

nathanbegbie:
Add typecasting to url_encode to handle non-string inputs

[Relevant slack thread here](https://turnio.slack.com/archives/C088C13PXNF/p1768824323502149)

Fix FunctionClauseError in URI.encode/2 when url_encode receives
non-string values like integers, floats, booleans, dates, or nil.

The issue occurred because URI.encode/2 only accepts binary strings,
but url_encode was passing values directly without conversion.

Changes:
- Add to_string() conversion before calling URI.encode()
- Add doctests for integer, float, boolean, nil, and date inputs
- Add doctest for complex value with integer in __value__ key

Fixes Sentry issue 8ff2f63e31ee488fbe739ce8780a6ad4

### DIFF
--- a/lib/expression/callbacks/standard.ex
+++ b/lib/expression/callbacks/standard.ex
@@ -2562,6 +2562,22 @@ defmodule Expression.Callbacks.Standard do
   @expression_category "string"
   @expression_doc expression: "url_encode(\"hello world\")",
                   result: URI.encode("hello world")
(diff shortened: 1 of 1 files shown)


smn (approved):
🙌  - great! thank you!

PR #306, opened 2026-01-21
Change release workflow to verify versions instead of auto-updating

nathanbegbie:
I tried to get smart by streamlining the release process as a single action (creating a Release on GitHub). Ended up looking like a fool!

CleanShot 2026-01-21 at 19 35 04@2x

Branch protection rules require verified signatures and PR-based changes, which GitHub Actions cannot satisfy.

So now we just fail faster instead of relying on hex and hexdocs.

I've also included a brief explanation of why we take this approach in the README.md in case someone else wants to try this in the future and will be forewarned of the hidden complexity there.

### DIFF
--- a/.github/workflows/releases.yml
+++ b/.github/workflows/releases.yml
@@ -14,13 +14,12 @@ jobs:
   publish:
     runs-on: ubuntu-latest
     permissions:
-      contents: write
+      contents: read
     steps:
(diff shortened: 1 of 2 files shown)


joaothallis (approved):
Looks good! Thanks for changes and documentation!!

PR #321, opened 2026-04-04
Phase 1: Safety and foundation

smn:
## Summary

- Replace `String.to_atom/1` with `String.to_existing_atom/1` in callback dispatch to prevent atom table exhaustion from arbitrary user input
- Add `%Expression.Context{}` struct with `vars` (expression-visible) and `private` (callback-only) fields, enabling private state isolation
- Add `Expression.Error` exception struct (`type`, `message`, `expression`, `position`) as the foundation for unified error handling

## Details

**Atom safety:** `atom_function_name/1` now uses `String.to_existing_atom/1` with a rescue fallback returning `nil`. `implements/3` restructured with `Code.ensure_compiled!` before atom lookup, nil-safe guards on all `function_exported?` calls. Vargs-only functions (e.g. `concatenate`) work correctly when the exact-name atom doesn't exist.

**Context struct:** Evaluator accepts both `%Expression.Context{}` structs and plain maps (dual-path). `Context.new/2` continues returning a plain map. New `Context.build/2` returns the struct with `private:` option. Variable resolution (`@foo`) reads from `vars` only; callbacks receive the full struct. Lambdas and captures work with structured context.

**Error struct:** Purely additive — no existing code uses it yet. Provides the type for Phases 3+.

## Test plan

- [x] Expression: 558 doctests + 255 tests, 0 failures
- [x] Engage: expression, callbacks, FLOIP, triggers tests — 0 failures
- [x] 8 new context struct tests verifying private state isolation

### DIFF
--- a/lib/expression/callbacks.ex
+++ b/lib/expression/callbacks.ex
@@ -34,7 +34,9 @@ defmodule Expression.Callbacks do
     do: atom_function_name("#{function_name}_")
 
   def atom_function_name(function_name) do
-    String.to_atom(function_name)
+    String.to_existing_atom(function_name)
+  rescue
+    ArgumentError -> nil
   end
 
   @doc """
@@ -62,9 +64,6 @@ defmodule Expression.Callbacks do
   end
 
   def implements(module \\ Standard, function_name, arguments) do
-    exact_function_name = atom_function_name(function_name)
-    vargs_function_name = atom_function_name("#{function_name}_vargs")
-
     # Make sure the module supplied and the default module are compiled
     # & loaded before attempting to find out what functions it may
     # support as part of validation / implementation checks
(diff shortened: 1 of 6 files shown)


lib/expression/callbacks.ex
```diff
@@ -73,24 +72,41 @@ defmodule Expression.Callbacks do
...
+          function_exported?(Standard, exact_function_name, length(arguments) + 1) ->
         {:exact, Standard, exact_function_name, length(arguments) + 1}
 
-      # Check if it's been implemented to accept a variable amount of arguments
-      function_exported?(Standard, vargs_function_name, 2) ->
+      # Check if it's been implemented to accept a variable amount of arguments in Standard
+      not is_nil(vargs_function_name) and function_exported?(Standard, vargs_function_name, 2) ->
         {:vargs, Standard, vargs_function_name, 2}
 
       # Check if the wrong number of arguments was provided
```
santiagocardo, line 112: Should we do something like this to avoid the `credo:disable` comment?
```suggestion
  defp do_implements(module, _function_name, exact_function_name, _vargs_function_name, _arguments)
    when not is_nil(exact_function_name) and exact_function_name in @built_in_operators,
    do: {:exact, module, exact_function_name, 2}

  # Check if the exact function signature has been implemented
  defp do_implements(module, _function_name, exact_function_name, _vargs_function_name, arguments)
    when not is_nil(exact_function_name) and function_exported?(module, exact_function_name, length(arguments) + 1),
    do: {:exact, module, exact_function_name, length(arguments) + 1}

      # Check if it's been implemented to accept a variable amount of arguments
  defp do_implements(module, _function_name, _exact_function_name, vargs_function_name, _arguments)
    when not is_nil(vargs_function_name) and function_exported?(module, vargs_function_name, 2),
    do: {:vargs, module, vargs_function_name, 2}

  # Check if the exact function signature has been implemented in Standard
  defp do_implements(module, _function_name, exact_function_name, _vargs_function_name, arguments)
    when not is_nil(exact_function_name) and function_exported?(Standard, exact_function_name, length(arguments) + 1),
    do: {:exact, Standard, exact_function_name, length(arguments) + 1}

  # Check if it's been implemented to accept a variable amount of arguments in Standard
  defp do_implements(module, _function_name, _exact_function_name, vargs_function_name, _arguments)
    when not is_nil(vargs_function_name) and function_exported?(Standard, vargs_function_name, 2),
    do: {:vargs, Standard, vargs_function_name, 2}

  defp do_implements(module, function_name, exact_function_name, vargs_function_name, arguments)
    cond do
      # Check if the wrong number of arguments was provided
```

test/expression_context_test.exs
```diff
@@ -14,4 +14,63 @@ defmodule ExpressionContextTest do
...
+      assert "yes" == Expression.evaluate_as_string!("@visible", ctx)
+      assert "@number" == Expression.evaluate_as_string!("@number", ctx)
+    end
+
+    test "callbacks receive the full context struct" do
+      ctx = Expression.Context.build(%{name: "Jane"}, private: %{token: "secret"})
+      # Evaluating a function passes the struct to the callback module
+      # The standard callbacks work with both maps and structs
+      result = Expression.evaluate_as_string!("@(upper(name))", ctx)
+      assert result == "JANE"
```
santiagocardo:
```suggestion
      assert Expression.evaluate_as_string!("@(upper(name))", ctx) == "JANE"
```

test/expression_context_test.exs
```diff
@@ -14,4 +14,63 @@ defmodule ExpressionContextTest do
...
+      ctx = Expression.Context.build(%{name: "Jane"}, private: %{token: "secret"})
+      assert ^ctx = Expression.Context.new(ctx)
+    end
+
+    test "lambdas work with structured context" do
+      ctx =
+        Expression.Context.build(%{items: [1, 2, 3]}, private: %{secret: "hidden"})
+
+      result = Expression.evaluate!("@(map(items, &(&1 + 1)))", ctx)
+      assert result == [2, 3, 4]
```
santiagocardo:
```suggestion
      assert Expression.evaluate!("@(map(items, &(&1 + 1)))", ctx) == [2, 3, 4]
```

santiagocardo (dismissed):
It looks nice, thanks for these small improvements ✨

lib/expression/callbacks.ex
smn: Oh wait still need to apply this one.

smn: hm this won't work because `function_exported`? is not allowed in a guard it seems. This is also being tackled in the 2nd PR so I'm choosing to leave it as is for now @santiagocardo

smn:
ready again @santiagocardo -- will look at the credo issue in the follow-up PR

lib/expression/callbacks.ex
santiagocardo: Alright @smn, thanks for checking!

santiagocardo (approved):
LGMT

PR #322, opened 2026-04-04
Phase 2: Developer experience — defexpr macro

smn:
## Summary

- Add `defexpr` macro for callback definition with automatic argument evaluation
- Add compile-time function registration and consistency validation
- Add `@variadic true` attribute support (replaces `_vargs` suffix convention)
- Add `use Expression.Callbacks` composition options (`stdlib:`, `also:`)

Depends on #321

## Details

**`defexpr` macro:** Two forms — `defexpr foo(a, b) do` (no context) and `defexpr foo(a, b), ctx do` (with context). Arguments are auto-evaluated from AST via generated `eval!` calls. Reserved words handled transparently: `defexpr or_(a, b)` registers as expression name `:or`.

**Compile-time infra:** `@expression_function` accumulated attribute with `validate_expression_func!/3` ensuring all clauses agree on context usage. `@before_compile` generates `__expression_functions__/0`. `@variadic` read at caller's compile time via `Module.delete_attribute` inside `quote` (Lua pattern).

**Composition:** `stdlib: false` disables Standard fallback. `also: [ModuleA, ModuleB]` for multi-module dispatch via `handle_chain/4`. Default `use Expression.Callbacks` behavior unchanged.

## Test plan

- [x] 21 new tests in `test/expression_defexpr_test.exs`
- [x] Expression: 558 doctests + 276 tests, 0 failures
- [x] Engage: 301 tests, 0 failures

### DIFF
--- a/lib/expression/callbacks.ex
+++ b/lib/expression/callbacks.ex
@@ -365,6 +365,27 @@ defmodule Expression.Callbacks do
   Expression function names that are Elixir reserved words (`and`, `if`,
   `or`, `not`) are automatically suffixed with `_` in the generated
   function name.
+
+  ## When NOT to use `defexpr`
+
+  `defexpr` eagerly evaluates **all** arguments before the body runs.
+  This is wrong for functions that need lazy or short-circuit evaluation:
+
+    - `if_(condition, yes, no)` — must only evaluate the matching branch
+    - `or_(a, b)` — must stop at the first truthy value
+
+  These functions must use plain `def` with manual `eval!` calls to
+  control when each argument is evaluated:
+
+      def if_(ctx, condition, yes, no) do
+        if eval!(condition, ctx),
+          do: eval!(yes, ctx),
+          else: eval!(no, ctx)
+      end
+
+  As a rule: if the function might skip evaluating some arguments
+  depending on the value of others, use `def`. If all arguments are
+  always needed, use `defexpr`.
   """
   defmacro defexpr(function_head, ctx_var, rest) do
     define_expr(function_head, ctx_var, rest)
--- a/lib/expression/callbacks.ex
+++ b/lib/expression/callbacks.ex
@@ -144,7 +144,8 @@ defmodule Expression.Callbacks do
 
     cond do
       not is_nil(exact) and exact in @built_in_operators ->
-        {:ok, apply(Kernel, exact, arguments |> Enum.map(&Expression.Eval.eval!(&1, context)))}
+        evaluated_args = Enum.map(arguments, &Expression.Eval.eval!(&1, context))
+        {:ok, apply(Kernel, exact, evaluated_args)}
 
       not is_nil(exact) and function_exported?(mod, exact, length(arguments) + 1) ->
         {:ok, apply(mod, exact, [context] ++ arguments)}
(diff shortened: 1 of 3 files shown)


lib/expression/callbacks.ex
```diff
@@ -130,12 +224,247 @@ defmodule Expression.Callbacks do
...
+  """
+  defmacro defexpr(fa, ctx_var, rest) do
+    define_expr(fa, ctx_var, rest)
+  end
+
+  defmacro defexpr(fa, rest) do
+    define_expr(fa, nil, rest)
+  end
+
+  defp define_expr(fa, ctx_var, do: body) do
```
santiagocardo: @smn, what id `fa`? Can we give it a more descriptive name instead?

lib/expression/callbacks.ex
```diff
@@ -130,12 +224,247 @@ defmodule Expression.Callbacks do
...
+    {name, args} = decompose_fa(fa)
+    with_ctx? = ctx_var != nil
+
+    # The Elixir def name is what the user writes (e.g., or_, if_, and_, not_)
+    def_name = name
```
santiagocardo: I think this is unnecessary, we can just call it `def_name` directly:
```suggestion
    # The Elixir def name is what the user writes (e.g., or_, if_, and_, not_)
    {def_name, args} = decompose_fa(fa)
    with_ctx? = ctx_var != nil
```

lib/expression/callbacks.ex
```diff
@@ -63,6 +90,73 @@ defmodule Expression.Callbacks do
...
+  defp find_in_chain([], function_name, _exact, _vargs, _arguments, _opts) do
+    {:error, "#{function_name} is not implemented."}
+  end
+
+  defp find_in_chain([mod | rest], function_name, exact, vargs, arguments, opts) do
+    context = Keyword.fetch!(opts, :context)
+
+    cond do
+      not is_nil(exact) and exact in @built_in_operators ->
+        {:ok, apply(Kernel, exact, arguments |> Enum.map(&Expression.Eval.eval!(&1, context)))}
```
santiagocardo:
```suggestion
        evaluated_args = Enum.map(arguments, &Expression.Eval.eval!(&1, context))
        {:ok, apply(Kernel, exact, evaluated_args)}
```

santiagocardo (approved):
Wow, this metaprogramming is 🤯 Also, really hard to follow, but I'm trusting the test suite!

lib/expression/callbacks.ex
smn: will update, `fa` is for "function application", ie the AST `{name, metadata, args}` tuple representation of a function

smn:
@santiagocardo spent more time on simplifying things & add documentation detailing how stuff works. Let me know if it's clearer now?

lib/expression/callbacks.ex
```diff
@@ -63,6 +90,120 @@ defmodule Expression.Callbacks do
...
+    |> Enum.uniq()
+    |> Enum.each(fn mod ->
+      Code.ensure_compiled!(mod)
+      Code.ensure_loaded!(mod)
+    end)
+
+    if is_nil(exact_function_name) and is_nil(vargs_function_name) do
+      {:error, "#{function_name} is not implemented."}
+    else
+      find_in_chain(
```
santiagocardo: Can we rename this to something like `apply_first_in_chain`?

lib/expression/callbacks.ex
```diff
@@ -123,19 +290,413 @@ defmodule Expression.Callbacks do
...
+    # `Module.delete_attribute(__MODULE__, :variadic)` reads the value
+    # AND clears it in one step (so it doesn't leak to the next defexpr).
+    # If @variadic was set, it returns `true` and we emit the variadic
+    # function. Otherwise we emit the exact-arity function.
+    #
+    # This pattern comes from the Lua package's `deflua` macro.
+    #
+    quote do
+      if Module.delete_attribute(__MODULE__, :variadic) do
+        unquote(variadic_ast)
```
santiagocardo: Can we maybe do this to compute the variadic function generation only when needed?

```suggestion
        unquote(gen_variadic_def(vargs_def_name, expr_name, user_args_var, ctx_var, with_ctx?, body))
```

lib/expression/callbacks.ex
```diff
@@ -123,19 +290,413 @@ defmodule Expression.Callbacks do
...
+    [quote(do: unquote(ctx_var) = unquote(ctx_param))]
+  end
+
+  # ── Compile-time validation ───────────────────────────────────────
+
+  @doc false
+  def validate_expression_func!({name, with_ctx?, variadic?}, module, existing) do
+    conflict =
+      Enum.find(existing, fn
+        {^name, other_ctx?, _} -> other_ctx? != with_ctx?
```
santiagocardo:
```suggestion
        {^name, other_with_ctx?, _} -> other_with_ctx? != with_ctx?
```

santiagocardo:
Yay, this detailed documentation is 💜 That's much better indeed. Thanks @smn!

lib/expression/callbacks.ex
santiagocardo: Thank you!!

PR #324, opened 2026-04-04
Phase 4: Parser improvements — infix and/or/not, ~EXPR sigil

smn:
## Summary

- Add infix `and`/`or`/`not` operator support with proper precedence
- Add `~EXPR` sigil for compile-time expression validation and optional pre-parsing

Depends on #323

## Details

**Infix logical operators:** New precedence chain: `aexpr` (or) → `aexpr_and` → `aexpr_not` (prefix unary) → `aexpr_compare`. Infix `a and b` produces `{:function, [name: "and", args: [a, b]]}` — identical AST to `and(a, b)` function call. Operators use `lookahead_not` on word characters to avoid matching inside `android`/`order`/`nothing`. Existing function call syntax unchanged.

**`~EXPR` sigil:** `Expression.Sigil` module. Default mode validates syntax at compile time, returns string. `c` modifier returns pre-parsed AST via `Macro.escape` — zero runtime parse cost. Invalid syntax raises `CompileError`.

## Test plan

- [x] 22 new tests for infix operators (basic ops, variables, comparisons, precedence, partial-word safety, backwards compat)
- [x] 6 new tests for `~EXPR` sigil
- [x] Expression: 558 doctests + 304 tests, 0 failures
- [x] Engage: 301 tests, 0 failures

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #325, opened 2026-04-04
Phase 5: Context normalization options

smn:
## Summary

- Add `lowercase_keys` and `coerce_strings` options to `Context.new/2`
- Add case-insensitive variable lookup in evaluator
- `skip_context_evaluation?` kept as backwards-compatible alias

Depends on #324

## Details

**Context options:** `lowercase_keys: false` preserves original key casing (atom keys still stringified). `coerce_strings: false` preserves all string values as-is. Default behavior (both `true`) unchanged. Options pass through `build/2`.

**Case-insensitive lookup:** `Eval.case_insensitive_get/2` tries exact match first (fast path), falls back to case-insensitive key scan. Enables `lowercase_keys: false` without breaking expressions — the parser already lowercases variable names in the AST, so `@firstname` resolves against a `"FirstName"` key.

**Engage audit:** Engage already uses lowercase string keys in context maps. No engage changes needed — options are opt-in for consumers with case-sensitive external data.

## Test plan

- [x] 18 new context tests (options, case-insensitive lookup, combined options, build passthrough)
- [x] Expression: 558 doctests + 320 tests, 0 failures
- [x] Engage: 301 tests, 0 failures

### DIFF
--- a/test/expression/context_test.exs
+++ b/test/expression/context_test.exs
@@ -76,6 +76,131 @@ defmodule Expression.ContextTest do
     end
   end
 
+  describe "lowercase_keys option" do
+    test "default: keys are lowercased" do
+      assert %{"name" => "Jane"} = Context.new(%{Name: "Jane"})
+    end
+
+    test "lowercase_keys: true is the same as default" do
+      assert %{"name" => "Jane"} = Context.new(%{Name: "Jane"}, lowercase_keys: true)
+    end
+
+    test "lowercase_keys: false preserves original key casing" do
+      ctx = Context.new(%{"FirstName" => "Jane", "lastName" => "Doe"}, lowercase_keys: false)
+      assert ctx["FirstName"] == "Jane"
+      assert ctx["lastName"] == "Doe"
+      refute Map.has_key?(ctx, "firstname")
+    end
+
+    test "lowercase_keys: false still converts atom keys to strings" do
+      ctx = Context.new(%{Name: "Jane"}, lowercase_keys: false)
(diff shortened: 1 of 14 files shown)


test/expression/context_test.exs
```diff
@@ -76,6 +76,131 @@ defmodule Expression.ContextTest do
...
+
+    test "coerce_strings: false preserves zero as string" do
+      ctx = Context.new(%{"zero" => "0"}, coerce_strings: false)
+      assert ctx["zero"] == "0"
+    end
+
+    test "coerce_strings: false preserves datetime strings" do
+      ctx = Context.new(%{"date" => "2020-12-13T23:34:45"}, coerce_strings: false)
+      assert ctx["date"] == "2020-12-13T23:34:45"
+    end
```
santiagocardo, line 161: Can we also add a test with `coerce_strings: true`?

santiagocardo (dismissed):
Super nice 🔥

smn:
Will look at merging this later. The earlier PRs don't seem to have merged into develop for some reason

smn:
Sorry @santiagocardo this now has PRs 2, 3, and 4 merged in because somehow I managed to screw up the PR chain. Apologies. You already reviewed & approved this earlier so hoping you can still follow along.

test/expression/context_test.exs
smn: good idea, added in 08bfb00

santiagocardo (approved):
LGTM!

PR #326, opened 2026-04-04
Phase 6: Cleanup — deprecate V2.Compat, document AST

smn:
## Summary

- Deprecate `Expression.V2.Compat` functions with `@deprecated` attributes
- Add `Expression.AST` module with full typespecs for all AST node types

Depends on #325

## Details

**V2.Compat deprecation:** `@deprecated` on `evaluate_as_string!/3`, `evaluate!/3`, `evaluate_block!/4`. Points to `Expression` equivalents. Moduledoc updated. Cannot remove yet — `flow_runner` uses it (all 136 flow_runner tests pass with deprecation warnings).

**AST documentation:** New `Expression.AST` module with typespecs for: `literal`, `variable`, `attribute`, `key`, `function_call`, `binary_op`, `list_node`, `range`, `lambda`, `capture`. Top-level `template` and `block` types. Full operator precedence documentation. Examples for every AST shape. Formalizes what engage's `ElasticsearchConverter` already depends on.

**Deferred:** Standard callbacks split into category modules — 4000+ line refactor with limited immediate value.

## Test plan

- [x] Expression: 558 doctests + 320 tests, 0 failures
- [x] Flow_runner: 136 tests, 0 failures (deprecation warnings visible)
- [x] Engage: expression + FLOIP tests, 0 failures

### DIFF
--- a/lib/expression/ast.ex
+++ b/lib/expression/ast.ex
@@ -1,0 +1,186 @@
+defmodule Expression.AST do
+  @moduledoc """
+  Type definitions for the Expression abstract syntax tree.
+
+  The AST is produced by `Expression.Parser` and consumed by `Expression.Eval`.
+  It is also used by consumers for transpilation (e.g., to Elasticsearch queries)
+  and static analysis (e.g., function validation). This module formalizes the
+  AST as a public API.
+
+  ## Top-Level Structure
+
+  `Expression.Parser.parse/1` returns a keyword list of `:text` and `:expression`
+  entries:
+
+      [text: "hello ", expression: [atom: "world"]]
+
+  `Expression.parse_expression!/1` returns the inner expression AST without
+  the `:text`/`:expression` wrappers.
+
+  ## Node Types
+
+  ### Literals
+  `{:literal, value}` — a constant value (integer, float, string, boolean, Date, DateTime, Time).
+
+      {:literal, 42}
+      {:literal, "hello"}
+      {:literal, true}
+      {:literal, ~D[2022-01-31]}
+
+  ### Variables
+  `{:atom, name}` — a variable reference. Names are always lowercased strings.
+
+      {:atom, "contact"}
+      {:atom, "name"}
+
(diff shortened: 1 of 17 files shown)


lib/expression/ast.ex
```diff
@@ -0,0 +1,186 @@
...
+  """
+
+  @typedoc "A literal value"
+  @type literal :: {:literal, number | String.t() | boolean | Date.t() | DateTime.t() | Time.t()}
+
+  @typedoc "A variable reference (always lowercased)"
+  @type variable :: {:atom, String.t()}
+
+  @typedoc "Dot-notation property access"
+  @type attribute :: {:attribute, list(ast_node)}
```
santiagocardo, line 127: @smn, can this be a bit more precise, saying it would actually just be something like this?
```suggestion
  @type attribute :: {:attribute, [variable] | [{:attribute, [variable]}]}
```

lib/expression/ast.ex
```diff
@@ -0,0 +1,186 @@
...
+
+  @typedoc "Bracket-notation index access"
+  @type key :: {:key, list(ast_node)}
```
santiagocardo, line 130: Same here

```suggestion
  @type key :: {:key, [variable | literal]}
```

santiagocardo (dismissed):
These improvements are 🚀!!

lib/expression/ast.ex
smn: I looked into this but the parser does allow nested expressions in these positions. This can be `@foo.123.bar` for example

lib/expression/ast.ex
smn: Same here, this could be `@items[count("foo")]` which means key values can be ast_nodes() which cover what you're suggesting here.

smn:
@santiagocardo fixed the underlying issues preventing a merge to `develop` -- could you approve again?

santiagocardo (approved):

smn:
thank you!

PR #327, opened 2026-04-04
Phase 7: Add evaluate_template/3

smn:
## Summary

- Add `Expression.evaluate_template/3` as the explicit, named function for resolving `@variable` references in strings

Depends on #326

## Details

`Expression.evaluate_template/3` delegates to `evaluate_as_string!/3`. This is the explicit alternative to the implicit behavior where `Eval.eval!({:literal, binary}, ...)` re-parses string literals as expression templates.

The implicit behavior remains unchanged for backwards compatibility. Consumers can migrate to the explicit function over time. Removing the implicit path is deferred to a future major version.

Remaining Phase 7 items (major-version-only, not actionable now):
- Flip `coerce_strings` and `lowercase_keys` defaults to `false`
- Remove nested literal re-evaluation implicit path

## Test plan

- [x] Expression: 560 doctests + 320 tests, 0 failures (2 new doctests from evaluate_template examples)
- [x] Engage: expression + FLOIP tests, 0 failures

### DIFF
--- a/lib/expression.ex
+++ b/lib/expression.ex
@@ -232,17 +232,19 @@ defmodule Expression do
   as an argument to a function). Use this when you need template resolution
   and want to be explicit about it.
 
+  Raises `Expression.Error` on parse or evaluation failures.
+
   ## Examples
 
-      iex> Expression.evaluate_template("hello @name", %{"name" => "world"})
+      iex> Expression.evaluate_template!("hello @name", %{"name" => "world"})
       "hello world"
 
-      iex> Expression.evaluate_template("1 + 1 = @(1 + 1)", %{})
+      iex> Expression.evaluate_template!("1 + 1 = @(1 + 1)", %{})
       "1 + 1 = 2"
 
   """
-  @spec evaluate_template(String.t(), map(), module()) :: String.t()
-  def evaluate_template(template, context \\ %{}, mod \\ Expression.Callbacks) do
+  @spec evaluate_template!(String.t(), map(), module()) :: String.t()
+  def evaluate_template!(template, context \\ %{}, mod \\ Expression.Callbacks) do
     evaluate_as_string!(template, context, mod)
   end
 
--- a/lib/expression.ex
+++ b/lib/expression.ex
@@ -223,6 +223,29 @@ defmodule Expression do
       "__value__" => nil
     }
 
+  @doc """
+  Evaluate a string as an expression template, resolving any `@variable`
+  references and `@(expression)` blocks within it.
+
+  This is the explicit version of the behavior that occurs implicitly when
+  a string literal appears inside an expression (e.g., `"hello @name"`
+  as an argument to a function). Use this when you need template resolution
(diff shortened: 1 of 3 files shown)


lib/expression/callbacks/standard.ex
```diff
@@ -48,7 +48,8 @@ defmodule Expression.Callbacks.Standard do
                     }
                   },
                   result: 23
-  def(count(ctx, term)) do
+
```
santiagocardo:
```suggestion
```

lib/expression.ex
```diff
@@ -223,6 +223,29 @@ defmodule Expression do
...
+
+      iex> Expression.evaluate_template("hello @name", %{"name" => "world"})
+      "hello world"
+
+      iex> Expression.evaluate_template("1 + 1 = @(1 + 1)", %{})
+      "1 + 1 = 2"
+
+  """
+  @spec evaluate_template(String.t(), map(), module()) :: String.t()
+  def evaluate_template(template, context \\ %{}, mod \\ Expression.Callbacks) do
```
santiagocardo: This should also be a bang function, no?
```suggestion
  @spec evaluate_template!(String.t(), map(), module()) :: String.t()
  def evaluate_template!(template, context \\ %{}, mod \\ Expression.Callbacks) do
```

santiagocardo (approved):
@smn, what would be a practical use case for this new function? I see this is just a wrapper function for `evaluate_as_string!/3`.

Also, can we add some tests for it?

lib/expression/callbacks/standard.ex
smn: the formatter is doing something odd here, if I remove the linebreak it formats the `def count(...)` as `def(count(...))`

smn: hm looks like that's now resolved, not sure why but thanks!

smn:
@santiagocardo I added it to address the literal re-evaluation problem here:

```elixir
  def eval!({:literal, literal}, context, mod) when is_binary(literal) do
    Expression.evaluate_as_string!(literal, context, mod)
  end
```

Every string literal inside an expression is implicitly re-parsed as a template. This is undocumented and can cause unintended recursion on string inputs. It also means you can't have a literal string containing @ without it being treated as a variable.

the proposed `evaluate_template/3` is the explicit alternative. The intended roadmap here is:

1. Add `evaluate_template/3` as the named function (this PR)
2. Future: Add a `nested_literal_evaluation: false` option that disables the implicit re-parsing in `eval!`
3. Major version: Flip the default to false — string literals stay as strings unless you explicitly call `evaluate_template/3`

It's the first step toward removing the implicit behavior without breaking anything today. Folks who rely on nested template resolution can migrate to calling `evaluate_template/3` explicitly.

santiagocardo:
Ah, ok @smn, now I get it! Makes sense to gradually introduce this change 👍

smn:
merging this into the target PR as it's minimal

PR #338, opened 2026-05-07
v3.0.0-rc.0: breaking changes

smn:
## Summary

Release candidate for Expression v3.0.0. Removes the entire V2 parser/evaluator, changes `Context.new/2` defaults to stop silently coercing values, fixes downstream-breaking bugs in the evaluator and error handling, and adds type coercion callback functions.

- **V2 removal:** All modules under `lib/expression/v2/`, benchmark scripts, `benchee` dep, and all V2 tests deleted (~3,300 lines removed)
- **Context defaults changed:** `lowercase_keys` and `coerce_strings` both default to `false`; callers must opt in to v2-style normalization
- **Structured error returns:** `evaluate_block/3` and `evaluate/3` now return `{:error, %Expression.Error{}}` instead of `{:error, string}`, so consumers can distinguish parse/eval errors from valid string values
- **Restored recursive literal evaluation:** String literals containing `@variable` references (e.g. `"@row.text"`) are again resolved at eval time, fixing breakage in downstream flow runner / Engage journeys
- **List index coercion:** `read_key_from_subject/2` now handles string indices on lists (`items["5"]` → 6th element) and float indices (`items[5.0]` → truncated to 5)
- **New type coercion functions:** `int()`, `number()`, `text()`, `boolean()` added to `Expression.Callbacks.Standard`
- **DateTime equality fix:** `=` on two `DateTime` values now compares the full DateTime (v2 incorrectly compared only the date portion)

### `Context.new/2` defaults

`lowercase_keys` and `coerce_strings` both default to `false`. String values pass through untouched and key casing is preserved (variable lookup remains case-insensitive at eval time). To restore v2 behavior:

```elixir
Context.new(ctx, lowercase_keys: true, coerce_strings: true)
```

The `:skip_context_evaluation?` option has been removed — use `coerce_strings: false`.

### Error return type

`evaluate_block/3` and `evaluate/3` now return `{:error, %Expression.Error{}}` instead of `{:error, string}`. The `Expression.Error` struct has `:type`, `:message`, `:expression`, and `:position` fields. Consumers pattern-matching on `{:error, string}` must update:

```elixir
# Before
{:error, "Unable to parse block: ..."} -> handle_error()

# After
{:error, %Expression.Error{message: msg}} -> handle_error()
```

Dead `RuntimeError` rescue clauses removed from both functions (the bang variants already wrap `RuntimeError` → `Expression.Error`).

### `Expression.error/1` removed

Use `Expression.error_map/1` for the legacy error map shape, or raise `Expression.Error`.

### V2 parser/evaluator removed

All modules deleted: `Expression.V2`, `V2.Parser`, `V2.Compile`, `V2.Context`, `V2.Eval`, `V2.Callbacks`, `V2.Callbacks.Standard`, `V2.Autodoc`, `V2.Compat`. Consumers using V2 for static analysis must migrate to V1's AST surface.

### DateTime `=` operator

`=` on two `DateTime` values now compares the full DateTime, matching `==`. v2 incorrectly compared only the date portion via `Date.compare/2`.

## Bug fixes

- **Recursive literal evaluation restored:** The `eval!({:literal, literal}, context, mod) when is_binary(literal)` clause (originally from `adbe1fa`) was accidentally removed during V2 deletion. Restored so that string literals containing `@variable` references resolve correctly. Without this, expressions like `"@row.text"` in flow runner Case blocks and `update_contact` calls return the literal `@row.text` string instead of the variable value.
- **String index on lists:** `items[contact.field]` where `contact.field` is `"5"` (string from DB) now coerces to integer and returns the correct element instead of `nil`.
- **Float index on lists:** `items[5.0]` now truncates to `5` instead of crashing (`Access.at` requires integer).

## New features

Four type coercion functions added to `Expression.Callbacks.Standard`:

| Function | Description | Examples |
|----------|-------------|---------|
| `INT(value)` | Convert to integer | `INT("5")` → `5`, `INT(5.7)` → `5`, `INT(true)` → `1` |
| `NUMBER(value)` | Convert to float | `NUMBER("5.5")` → `5.5`, `NUMBER(5)` → `5.0` |
| `TEXT(value)` | Convert to string | `TEXT(5)` → `"5"`, `TEXT(nil)` → `""` |
| `BOOLEAN(value)` | Convert to boolean | `BOOLEAN(0)` → `false`, `BOOLEAN("hello")` → `true` |

These enable explicit type coercion in expressions, e.g. `rows[INT(contact.lastmathworksheetsent)]` instead of the `parse_json()` workaround.

## Other changes

- Callbacks cleanup: removed unnecessary `is_binary` / `is_list` / `is_nil` guards from `regex_capture`, `regex_named_capture`, `with_index`, `uniq`, `has_all_members`, `has_any_member`, `has_any_phrase`
- Doctest context values updated to use proper types instead of strings (where `coerce_strings` was implicitly relied on)
- `error_map/1` promoted from `@doc false` to a documented public function
- Version bumped to `3.0.0-rc.0`
- `benchee` dependency removed

## Test plan

- [x] `mix test` — 413 doctests, 240 tests, 0 failures
- [x] `mix credo --strict` — no issues
- [x] `mix format --check-formatted` — clean
- [ ] Run engage CI against `3.0.0-rc.0` to surface migration work
- [ ] Address engage fallout (V2 deps in `Turn.Dsl.Validation` / `Turn.Dsl.Parser`, error tuple consumers in flow runner)
- [ ] Tag `3.0.0` final once rc is green

### flow_runner

`FlowRunner.evaluate_expression_block/2` currently unwraps `{:error, reason}` into a plain string. With the new `{:error, %Expression.Error{}}` return, update to:

```elixir
case Expression.evaluate_block(expression, context, expression_callbacks_module()) do
  {:ok, val} -> val
  {:error, %Expression.Error{}} -> {:error, :parse_failure}
end
```

### engage

- Replace any `Expression.V2` / `V2.Parser` / `V2.Compile` usage with V1 equivalents
- Replace `Expression.error/1` with `Expression.error_map/1`
- Update any `{:error, string}` pattern matches on `evaluate_block/3` / `evaluate/3` results

## Files

See `CHANGELOG.md` for the user-facing changelog and `V3_PLAN.md` for the implementation plan.


lib/expression/eval.ex
```diff
@@ -205,7 +201,7 @@ defmodule Expression.Eval do
     do: DateTime.compare(a, b) == :eq
 
   def op(:=, a, b) when is_struct(a, DateTime) and is_struct(b, DateTime),
-    do: Date.compare(a, b) == :eq
+    do: DateTime.compare(a, b) == :eq
```
santiagocardo, line 215: I believe that at some point should highlight this as a syntax error. Users should use `==` instead.

CLAUDE.md
```diff
@@ -40,12 +40,11 @@ mix dialyzer --format dialyxir
...
-### Two Parser Implementations
-The library maintains both V1 and V2 parsers for compatibility:
-- **V1** (`lib/expression/parser.ex`, `lib/expression/eval.ex`) - Original implementation
-- **V2** (`lib/expression/v2/`) - An EXPERIMENTAL and SOON TO BE THROWN AWAY parser with `compile.ex`, `eval.ex`, `context.ex`. MAKE NO CHANGES TO V2
+The parser and evaluator live under `lib/expression/parser.ex` and
+`lib/expression/eval.ex`. Parsing uses
+[NimbleParsec](https://hexdocs.pm/nimble_parsec).
 
-Both use [NimbleParsec](https://hexdocs.pm/nimble_parsec) for parsing.
+(The experimental V2 parser/evaluator was removed in v3.0.0.)
```
santiagocardo, line 47: @smn, if there are no more references to V2, we can remove this to avoid confusion, no?

santiagocardo (dismissed):
This is great 🔥✨🎉
Happy to try this new version on PROD. I hope this fixes existing bugs, especially the ones related to type coercions 🤞

lib/expression/eval.ex
smn: good point - do we do this now? This is a breaking change and so perhaps now is the time? What would `=` do given we don't do assignments?

santiagocardo: I believe we need to keep it until we migrate to v3 completely on Engage's side. Otherwise, this could break existing expressions in PROD. We should highlight this as a syntax error in the DSL first.

Also, in the DSL, we do assignments, but only in the body; these should not be allowed as a comparison operator.

Don't do ❌
```elixir
is_allowed = if contact.opted_in_at = now(), do: true, else: false
```

Do ✅
```elixir
is_allowed = if contact.opted_in_at == now(), do: true, else: false
```

smn: Yeah that assignment is actually handled in the FLOIP handling, not in the expression library. Everything to the right of the assignment is the expression being evaluated and nothing there is doing true variable assignments

smn:
@santiagocardo ready for review again

lib/expression/callbacks/standard.ex
```diff
@@ -2316,6 +2316,213 @@ defmodule Expression.Callbacks.Standard do
...
+                  expression: "int(value)",
+                  context: %{
+                    "value" => %{
+                      "display" => "value for display key",
+                      "value" => "value for value key",
+                      "__value__" => "42"
+                    }
+                  },
+                  result: 42
+  def int(ctx, value) do
```
santiagocardo, line 2386: @smn, let's rather rename this to something like: `to_int()`

lib/expression/callbacks/standard.ex
```diff
@@ -2316,6 +2316,213 @@ defmodule Expression.Callbacks.Standard do
...
+                  expression: "number(value)",
+                  context: %{
+                    "value" => %{
+                      "display" => "value for display key",
+                      "value" => "value for value key",
+                      "__value__" => "3.14"
+                    }
+                  },
+                  result: 3.14
+  def number(ctx, value) do
```
santiagocardo, line 2429: Same here? `to_number()`

lib/expression/callbacks/standard.ex
```diff
@@ -2316,6 +2316,213 @@ defmodule Expression.Callbacks.Standard do
...
+                  expression: "text(value)",
+                  context: %{
+                    "value" => %{
+                      "display" => "value for display key",
+                      "value" => "value for value key",
+                      "__value__" => 42
+                    }
+                  },
+                  result: "42"
+  def text(ctx, value) do
```
santiagocardo, line 2467:
```suggestion
  def to_text(ctx, value) do
```

lib/expression/callbacks/standard.ex
```diff
@@ -2316,6 +2316,213 @@ defmodule Expression.Callbacks.Standard do
...
+                  expression: "boolean(value)",
+                  context: %{
+                    "value" => %{
+                      "display" => "value for display key",
+                      "value" => "value for value key",
+                      "__value__" => 1
+                    }
+                  },
+                  result: true
+  def boolean(ctx, value) do
```
santiagocardo, line 2512:
```suggestion
  def to_boolean(ctx, value) do
```

santiagocardo (dismissed):
LGTM 🚀 Let's just agree on the new expression naming!

santiagocardo:
@smn, do you think we should go with `3.0.0-rc.0` or `3.0.0` directly? So I can use that version on `flow_runner`.

smn:
Either will work as a semantic version number. It's up to you!

Djack1010 (approved):

joaothallis (approved):

PR #341, opened 2026-05-20
[Callbacks] Guard against nil arguments to prevent Protocol.UndefinedError (PROD-5800)

smn:
## Purpose

Several callback functions in `Expression.Callbacks.Standard` crash with `Protocol.UndefinedError` when arguments resolve to `nil` at runtime. This happens in journey expressions when users reference unset contact fields or variables that evaluate to `nil`.

The original crash was `Enum.map(nil, ...)` inside `has_any_phrase/3`, but an audit revealed 7 additional functions with the same vulnerability.

### Link to the story

PROD-5800

## Approach

Added type guards to 8 callback functions, following the same pattern already used by safe sibling functions (`filter`, `find`, `map`, `reduce`, `reject`, `sort_by`):

| Function | Guard | Nil behavior |
|---|---|---|
| `has_any_phrase/3` | `cond` with `is_binary`/`is_list`/fallback | returns `false` |
| `has_all_members/3` | `is_list(list) and is_list(items)` | returns `false` |
| `has_any_member/3` | `is_list(list) and is_list(items)` | returns `false` |
| `has_any_word/3` | `to_string(words)` before `String.split` | returns `false` |
| `with_index/2` | `is_nil \|\| !Enumerable.impl_for` | returns `[]` |
| `uniq/2` | `is_nil \|\| !Enumerable.impl_for` | returns `[]` |
| `regex_capture/3` | `is_binary(binary)` | returns `nil` |
| `regex_named_capture/3` | `is_binary(binary)` | returns `%{}` |

## Security & Privacy

No security or privacy impact. This change only adds defensive guards that prevent crashes — no new data access or exposure.

## Learning

Followed the existing nil-guard patterns in the codebase (e.g. `filter/3` at line 591) for consistency.

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -19,7 +19,7 @@ jobs:
       - name: "CLA Assistant"
         if: (github.event.comment.body == 'recheck' || github.event.comment.body == 'I have read the CLA Document and I hereby sign the CLA') || github.event_name == 'pull_request_target'
         # Beta Release
-        uses: cla-assistant/github-action@v2.6.1
+        uses: cla-assistant/github-action@ca4a40a7d1004f18d9960b404b97e5f30a505a08 # v2.6.1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           # the below token should have repo scope and must be manually added by you in the repository's secret
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -33,14 +33,14 @@ jobs:
         cache_version: ["3"]
 
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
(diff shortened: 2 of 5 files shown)


santiagocardo (approved):
Since sibling functions already utilize these guards, I am happy with this approach 👍

However, I have a slight concern that it might just propagate errors to subsequent functions/expressions rather than crashing early.

smn:
@santiagocardo I agree with the concern, the problem is, there's no good mechanism right now to handle low level errors within an expression. Thoughts on how to handle that?

santiagocardo:
> @santiagocardo I agree with the concern, the problem is, there's no good mechanism right now to handle low level errors within an expression. Thoughts on how to handle that?

@smn, I don't think we should fix them directly; instead, we should log them so users are aware of these runtime errors. As done here:

https://github.com/turnhub/engage/blob/2857e6f05fd1efff60701d8f74b34036ac34fe3a/lib/turn/public_exceptions.ex#L186-L204

PR #370, opened 2026-08-12
fix: underscored number literals lose leading zeros of digit groups

fedme:
Expression supports Elixir-style underscore separators in number literals (`123_456`), but the `int()` parser captured each digit group between underscores with `integer/1`. Parsing `"000"` as an integer gives `0`, so `180_000` joined into `1800` and `1_000_000` into `100`. Elixir's code formatter writes any number longer than five digits with underscore separators, so expression strings generated from Elixir ASTs hit this constantly: `5000 > 180_000` evaluated to `true` and `200_000 + 1` to `2001`. Since the result is still a valid boolean or number, nothing ever logged.

The fix captures digit groups as strings so leading zeros survive the join, and gives floats the same underscore support since the formatter emits `1_234_567.89` for large floats too. Existing leniency is preserved: `1_` still parses as `1` and `1__1` still falls back to text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ by adding `expression` to your list of dependencies in `mix.exs`:
 ```elixir
 def deps do
   [
-    {:expression, "~> 3.0.0-rc.1"}
+    {:expression, "~> 3.0.0-rc.2"}
   ]
 end
 ```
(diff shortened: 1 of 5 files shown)


santiagocardo (dismissed):
Wow 🤯 I would never imagine this bug haha - thanks @fedme!

Could you please update `mix.exs` and `README.md` to `version: "3.0.0-rc.2"` to release it directly after merging?

fedme:
Done, bumped to 3.0.0-rc.2 in mix.exs and README.

santiagocardo (approved):

